PR #49, opened 2023-11-16
[FIX] certification_organism: fix demo hr leave conflict

vava-odoo:


xavieralt:
looks good, thanks for the quick fix :+1:

PR #55, opened 2023-11-28
[CLN] *: uniformize and clean data modules

vava-odoo:
- remove `color` field (random) definition
- uniformize `noupdate` status of (demo) data
- better manage product product/template
- remove editor history
- remove useless `studio` context
- remove industry name in xmlids definitions and references
- remove useless `model` on ir.model.fields if `model_id` is given
- define modules needed in manifest instead of config booleans
- avoid view inheritance from same module


bar_and_lounge/data/pos_config.xml
```diff
@@ -2,35 +2,30 @@
 <odoo noupdate="1">
     <record id="pos_config_2" model="pos.config">
         <field name="name">Porter (International Airport)</field>
-        <field name="module_pos_hr" eval="True" />
```
pimodoo, line 5: Good to have added the dependency, but you also need to keep it in pos config (pos config does not work the same way as res.config.settings)

PR #124, opened 2024-05-21
17.0 industry translation dhrs

dhrs-odoo:
Produce all translation file templates (POT files) for the industry modules.

Task-3734243


tsabi:
Finally :)

Because there is no issues enabled for this repository, i would like to mention here that the knowledge pages are not translatable. Can someone open a task inside odoo project management, that this could be a good thing to fix?

vava-odoo:
> Finally :)

Yes, with https://github.com/odoo/odoo/pull/165449 it is now possible to load translations for data modules.

> Because there is no issues enabled for this repository, i would like to mention here that the knowledge pages are not translatable. Can someone open a task inside odoo project management, that this could be a good thing to fix?

It makes sense, indeed, but I'm not aware if there is any mechanism for translation in knowledge articles. I'll investigate. Thanks

vava-odoo:
Hey @tsabi little update

> Because there is no issues enabled for this repository

Now there are.

> i would like to mention here that the knowledge pages are not translatable

We managed to translate them in 18 with the use of template that are rendered in the user language at installation.

Cheers!

PR #126, opened 2024-05-31
[FIX] software_reseller, corporate_gifts:  move Properties to new fields

dhrs-odoo:
In this PR we have replaced the task_properties_definition field to new fields.

Task-3942609


dhrs-odoo:
As this is cancelled so this PR is of no use 
So I am closing this PR

PR #140, opened 2024-06-14
[FIX] *: Records visible in demo not in data

jaeschwa:
Update all industry modules so all is_published are set to false
    in data and set to true in demo. The records are now only visible
    in demo and no more in data. Created related files for the different
    records.
    
    Added test to ensure that:
    -  all files in folder are in the manifest and vice versa
    -  Records with is "sale_ok" set to False are not with
       is_published="True"
    -  All knowledge.article records are not with is_published="True"
    -  The same field in a record is not updated twice in the same file.
    
    task-3945272


vava-odoo (commented):
Almost good. Looks like you forgot `hair_salon/demo/appointment_type.xml` :eyes:

vava-odoo (approved):

PR #141, opened 2024-06-14
[FIX] *: Records visible in demo not in data

jaeschwa:
Update all industry modules so all website_published are deleted
    in data and set to true in demo with the is_published set to True in
    data.
    
    Added test to ensure that:
    -  All files within the data folder does not have the argument
       website_published set to True
    -  No model should have the field website_published but rather the
       is_published argument.
    - Added a test so there are no duplicate field definition within the 
      same record

    
    task-3945272


industry_real_estate/data/website_controller_page.xml
```diff
@@ -104,7 +104,6 @@
         <field name="page_type">listing</field>
         <field name="model_id" ref="analytic.model_account_analytic_account"/>
         <field name="website_id" ref="website.default_website"/>
-        <field name="website_published" eval="True"/>
```
vava-odoo, line 107: This page should stay published. You'll need to update the test to allow this one.

tests/test_generic/tests/test_xml.py
```diff
@@ -431,6 +432,15 @@ def _check_is_published_false(self, xml_content, file_name):
         if '<record ' in xml_content and '<field name="is_published" eval="True"/>' in xml_content:
             _logger.warning(f"Model in {file_name} should not have 'is_published' set to True.")
 
+    def _check_website_published_false(self, xml_content, file_name):
+        if (
+            '<record ' in xml_content
+            and '<field name="website_published" eval="True"/>' in xml_content
+        ):
+            _logger.warning(
+                f"Model in {file_name} should not have 'website_published' set to True."
```
vava-odoo, line 441: Maybe you could be guiding a bit more, and provide a message mentioning that it is better to use `is_published` and only in demo (same comment for the test about is_published)

solar_installation/demo/im_livechat_channel.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo noupdate="1">
     <record id="im_livechat.im_livechat_channel_data" model="im_livechat.channel">
-        <field name="website_published" eval="True"/>
+        <field name="is_published" eval="True"/>
         <field name="is_published" eval="True"/>
```
vava-odoo, line 5: Defined twice :eyes: and your test doesn't raise...

vava-odoo (commented):
Thanks for the follow-up. I have a few comments, otherwise looks good.

solar_installation/demo/im_livechat_channel.xml
jaeschwa: I think the test does not raise because it is in the same definition, will add a new test for these cases

vava-odoo (approved):
Great. Just a little comment and you can squash everything into a single commit. Thanks!

PR #180, opened 2024-07-26
[FIX] *: do not post invoices in demo data 

dhrs-odoo:
Before this commit, the problem was that posting an invoice is blocked in the 
Indian localization if the company is not correctly configured. So when we
install any industry with demo data trying to post an invoice, the issue
occurs.

In this commit, we remove the action_post and associated payment methods
from all industries. This change ensures that the database remains clean and
free from issues when installing industries with demo data.

Task-4074862
opw-4006335


it_hardware/demo/purchase_order_confirm.xml
```diff
@@ -69,9 +65,6 @@
         <value model="account.move" eval="obj().env.ref('it_hardware.purchase_order_2').invoice_ids.ids"/>
         <value eval="{'invoice_date': DateTime.today().date()}"/>
```
vava-odoo, line 66: I think you can remove the invoice_date on purchase order invoices as well since you don't post it anymore.

vava-odoo (approved):
Thanks @dhrs-odoo 
I only have a single comment.
Ah, and it looks like you forgot one [here](https://github.com/odoo/industry/blob/b5248a37256cc79b37d56ebfbca7b1087271e64c/industry_real_estate/demo/account_move_confirm.xml#L3).
Cheers!

dhrs-odoo:
Hello @vava-odoo 
I have made the changes 
Thank You :innocent:

coal_petroleum/demo/purchase_order_post.xml
```diff
@@ -203,15 +203,6 @@
                 obj().env.ref('coal_petroleum.purchase_order_6')
             ).invoice_ids.ids
         "/>
-        <value eval="{'invoice_date': DateTime.today().date()}"/>
```
vava-odoo, line 206: If you remove that, the function becomes useless. You can therefore remove the whole `<function> `

dhrs-odoo:
Hello @vava-odoo 
I have updated your last comments changes
Thank You :innocent:

vava-odoo (approved):

PR #195, opened 2024-08-08
[FIX] *: Fixing the nonfunctionnal industries

jaeschwa:
Fixing the industries that are not working in master in the runbot.
This inclue the architect, real estaten industry restaurant , micro
brewery, sports club.

task-4103697

PR #196, opened 2024-08-13
[FIX] industry_restaurant: Table number instead of name

GaetanVandenBergh:
In restaurant table, the 'name' field become an integer field called table_number. The change made in this commit aims at conserving previous name that were integer. For the name that were not, the field is set to a unique integer higher than the previous maximum table number in the database.

Error seen in https://github.com/odoo/industry/pull/194
Caused by https://github.com/odoo/odoo/pull/173108


GaetanVandenBergh:
Hello @vava-odoo Not sure I need to go further, seems the module require more adaptation to be compatible with master
Latest build fail due to https://github.com/odoo/odoo/pull/165807/files

vava-odoo:
Thanks very much @GaetanVandenBergh but it is already handled in https://github.com/odoo/industry/pull/195 as there are other modules to fix.

PR #197, opened 2024-08-13
[FIX] industry_restaurant: improved ui for table booking

GaetanVandenBergh:
Caused by https://github.com/odoo/enterprise/pull/67182


GaetanVandenBergh:
Initial PR don't have the change https://github.com/odoo/enterprise/pull/67160

PR #215, opened 2024-09-11
[IMP] *: Renaming Industries and Categories

jaeschwa:
Renaming the different categories and the different modules to better
represent their range and utility.

task-4111596


jcs-odoo:
@jaeschwa Also rename "Automobile Spare Parts" (split "spareparts" into two words)

PR #220, opened 2024-09-12
[ADD] fast_food: Adding the fast_food industry

jaeschwa:
- Set up POS with "Kiosk" and "Drinks Display".
- Loaded "Restaurant" scenario.
- Added Loyalty Program.
- Imported res.partners, products, employees, planning, and purchases from Fine Dining.(including the replenishment and the supplier for certains product)
- Updated icon and image from

task-4069778


vava-odoo:
Will be merged in master with #230

PR #227, opened 2024-09-18
[IMP] *: knowledge links

jaeschwa:
Now the knwoledge links points to the correct version of the documentation

task-4116714


vava-odoo (changes requested):
I thought you'd make a test to ensure the version in the link is equivalent to the db version

PR #233, opened 2024-09-23
[FIX] test_xml: Fixing typo

jaeschwa:
In the test_xml.py file there was an issue the function : _check_is_published_false was not set on the xml content but on the modules. This commit aims to fix this issue.


tests/test_generic/tests/test_xml.py
```diff
@@ -34,7 +34,7 @@ def _check_files_in_path(self, module):
             for file_name in files:
                 file_path = os.path.join(root, file_name)
                 ext = os.path.splitext(file_path)[1].lower()
-                if 'static/' in file_path:
+                if 'static/' in file_path and 'src/js' not in file_path:
```
vava-odoo, line 37:
```suggestion
                if 'static/' in file_path:
```

vava-odoo (approved):

PR #244, opened 2024-09-26
[I18N] update .tx/config

vava-odoo:


ticodoo (approved):
Good thinking, totally forgot about this :see_no_evil:

PR #260, opened 2024-10-07
[ADD] .*: missing industries in the master

dhrs-odoo:
In this PR we have added the industries which are not available in the master

Task-4234855


vava-odoo:
Done in https://github.com/odoo/industry/pull/264

PR #262, opened 2024-10-08
[FIX] .*: fix the pot files for all the industries

dhrs-odoo:
In this PR we have export the pot files for all
the industries as per the latest changes.

Task-4216940


vava-odoo (approved):
Thanks @dhrs-odoo

PR #276, opened 2024-11-04
[FIX] headhunter: Sign request (#273)

jaeschwa:
Following this change : https://github.com/odoo/enterprise/pull/71542. Now the sign.request require the user to have valid email adresse. Also the sign request should be placed in demo and not in data. Moving the sign request in demo.


vava-odoo:
See https://github.com/odoo/industry/pull/281

PR #283, opened 2024-11-06
[FIX] beverage_distributor: Removing domain

jaeschwa:
In the module the domain was restricting the access of the product we can chose as subpart of another.

task-4243565


jaeschwa:
@vava-odoo , are you okay with removing this? Jean is functionally ok

vava-odoo (approved):

PR #310, opened 2024-11-26
[FIX] architects: fix expense validation mode

vava-odoo:
Since this commit*, there is a check on the bank account of the employee if the payment method is the default (own). This commit changes the payment mode to the 'company' one.

https://github.com/odoo/odoo/commit/2dc6b6c27176c51d879b149204b642a0324ea1cb


vava-odoo:
@robodoo r+

PR #315, opened 2024-11-27
[REV] *: replace ir.model.data by cloc_exclude entries in manifest"

vava-odoo:
This reverts commit 5802a3a13da3d3729ad283051b2ea13c74a6c0b9 because there is a bug in standard, preventing to upgrade data modules.


vava-odoo:
@robodoo fw=no

PR #317, opened 2024-11-27
[IMP] *: replace ir.model.data by cloc_exclude entries in manifest

vava-odoo:
In this commit, we replace all entries of ir.model.data by adding the specific files into the cloc_exclude list in the manifest. This will automatically create the necessary ir.model.data. Also, the demo files that could impact the cloc, mainly website views, are now excluded as well.

Task-4243531


vava-odoo:
@robodoo r+

PR #319, opened 2024-11-29
[FIX] *: forcecreate for standard records

vava-odoo:
Some standard records may have already been deleted before the installation of an industry module that relies on it. This commit turns on the `forcecreate` attribute on those records to avoid installation from failing.

opw-4277227


vava-odoo:
@robodoo r+ squash

PR #327, opened 2024-12-04
[ADD] condominium: add condominium in the attendees search

dhrs-odoo:
In this PR we have added a condominium in the attendees search so that we get all the owners belonging to this condominium.

task-4345839


vava-odoo (approved):
@robodoo r+

PR #329, opened 2024-12-05
[FIX] condominium: update pot for translations

vava-odoo:
Forgotten in #327


vava-odoo:
@robodoo r+
cc @dhrs-odoo

PR #334, opened 2024-12-09
[FIX] *: don't delete standard attributes

vava-odoo:
Before this commit, the standard product attributes were deleted at the installation of the industry module. The problem is that deletion can easily be forbidden, as soon as there is a product created with this
attribute.
This commit therefore removes the deletion of standard attributes. It also reorders attributes such that industry-specific ones are at the top.

*toy_store, micro_brewery, beverage_distributor

task-4393043


vava-odoo:
@robodoo r+

PR #337, opened 2024-12-10
[IMP] condominium: add automation to compute usage of meter readings

vava-odoo:
Before this commit, the usage of meter readings needs to be inserted manually. After this commit, the usage is computed on all meter readings related to the current property.


vava-odoo:
@robodoo r+

PR #340, opened 2024-12-17
[FIX] industry_real_estate: allow installation when crm stages are missing

hadielyakhni:
Since in 17.0, `forcecreate="True"` for creating other records's models is not supported, we bypass creating them for real_estate to skip the installation error.

opw-4277227


vava-odoo (approved):
@robodoo r+

PR #341, opened 2024-12-17
[FIX] *: forcecreate for standard records

vava-odoo:
Some standard records may have already been deleted before the installation of an industry module that relies on it.

This commit turns off the `forcecreate` attribute on those records to avoid installation from failing.

This reverts 198908503 as forcecreate="True" is not available in 17 for a record with an external id from another module.


vava-odoo:
@robodoo fw=no r+

PR #346, opened 2024-12-19
[FIX] *: upgrade is_favorite and display_in_project + auto_sequence

vava-odoo:
**[FIX]** *: **remove display_in_project on project.task**

The field `display_in_project` is now a readonly field, and should not be used to display some subtasks and not others. There is now a filter in the view to display sub-tasks.

Ref: odoo/odoo@46adf76

**[IMP] odoo_partner: auto_sequence tasks in project**

**[FIX] hair_salon: declare is_favorite on product.template**

The field `is_favorite` is now defined on product.template only, and cannot be changed on product.product

Ref: https://github.com/odoo/odoo/commit/41d8fa352a7baec

### DIFF
--- a/bar_industry/demo/project_task.xml
+++ b/bar_industry/demo/project_task.xml
@@ -14,28 +14,24 @@
             ]]></field>
         <field name="project_id" ref="project_project_1"/>
         <field name="stage_id" ref="project_task_type_1"/>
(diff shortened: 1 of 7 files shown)


vava-odoo:
@robodoo r+

vava-odoo:
@robodoo rebase-ff

PR #347, opened 2024-12-20
[FIX] fmcg_store: give valid module name

mega-odoo:
An error occurs due to give an invalid module name of `fmcg_store` at [1].

Link [1]: https://github.com/odoo/industry/blob/bb0f691d97bbe3fdd297d034da3b54f1cb59509b/fmcg_store/__manifest__.py#L65

Sentry-5974074704


vava-odoo (approved):
That was fixed in master but not in stable :man_facepalming: 
Thanks!
@robodoo r+
@robodoo fw=no

PR #361, opened 2025-01-15
[UPG] *: upgrade product packaging into uom

vava-odoo:
This commit removes product packaging and product category, as well as refactor the use of uom, following the change in standard*.

*see https://github.com/odoo/odoo/commit/d3fa388cf7c41984b5f7cbc2a81fec3e27a00325


vava-odoo:
@robodoo r+ rebase-ff

PR #362, opened 2025-01-16
[FIX] *: remove theme dependency

jaeschwa:
Steps to reproduce:
----

- Go on a DB without the design-themesµ
- Try to install a industries with design theme

Issue:
---

The indsutries were dependannt of the design themes (dependencies). Thus it was blocking users from installing them when the design themes were not installed.

Fix:
---

Removed the dependnencies since we are doing a search to find the dependencies.

task-4438105

### DIFF
--- a/agriculture_shop/demo/website.xml
+++ b/agriculture_shop/demo/website.xml
@@ -2,7 +2,7 @@
 <odoo noupdate="1">
     <record id="website.default_website" model="website">
         <field name="name">Agriculture</field>
-        <field name="theme_id" ref="base.module_theme_enark"/>
+        <field name="theme_id" search="[('id', '=', ref('base.module_theme_enark', raise_if_not_found=False))]"/>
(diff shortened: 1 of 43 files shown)


agriculture_shop/demo/website.xml
```diff
@@ -2,7 +2,6 @@
 <odoo noupdate="1">
     <record id="website.default_website" model="website">
         <field name="name">Agriculture</field>
-        <field name="theme_id" ref="base.module_theme_enark"/>
```
vava-odoo, line 5: I thought about it. Why don't we keep it, and simply don't raise if it is not there? So that we have the correct value if it is present. Something like
```xml
        <field name="theme_id" search="[('id', '=', ref('base.module_theme_enark', raise_if_not_found=False))]"/>
```

jaeschwa: Yes! You're right

vava-odoo:
@robodoo r+

PR #372, opened 2025-01-17
[FIX] art_craft: fix point of sale config

vava-odoo:
opw-4491888

backport of #361


vava-odoo:
@robodoo fw=no

vava-odoo:
@robodoo r+

PR #375, opened 2025-01-17
[IMP] construction, pharmacy_retail: remove blocking warnings

lugr-odoo:
Warnings set by the user are no longer blocking in sale, purchase and stock. Additionally, internal warnings were completely removed from accounting.

See odoo/odoo#192211


vava-odoo (approved):
Thanks very much for taking care of industries, @lugr-odoo 
@robodoo r+

PR #379, opened 2025-01-17
[FIX] electronic_store: fix categories in PoS

vava-odoo:
task-4485769


vava-odoo:
@robodoo fw=skipci

vava-odoo:
@robodoo r+

PR #387, opened 2025-01-21
[FIX] *: deposit product category instead of xmlid

vava-odoo:
*beverage_distributor, micro_brewery

Before this commit, the deposit category was the one created by the industry module. A product was defined as a deposit if its category was the same one. This methodology is quite limited as it relies on the ID of the category, leading to several potentiel issues (deletion, new additional categories, ...).

This commit adds a checkbox on the product category such that any product belonging to a category with this checkbox filled is marked as a deposit product.

task-4493015


vava-odoo:
@robodoo r+

vava-odoo:
@robodoo r+ rebase-ff

PR #388, opened 2025-01-23
[IMP] condominium: add button for condominium creation

jaeschwa:
This commit simplifies the condominium creation process by:

* Adding a "Create Condominium" button to the `res.partner` form.
* Added server action for the button to create a new condominium record.
* Setting visibility conditions for the button domain.
* Triggering a form refresh after creation to ensure visibility.

This improves the workflow for users by allowing them to create condominiums directly from the partner form.

task-4482863


vava-odoo (approved):
@robodoo r+

PR #389, opened 2025-01-23
[FIX] billboard_rental: don't edit subscriptions from billboard

vava-odoo:
When adding a subscription in the sub tab of a billboard form view, a traceback appears. As the goal of the tab is to display the linked subscriptions (and not create new ones), the field is now marked as readonly. 

task-4501934


vava-odoo:
@robodoo squash
@robodoo fw=no
@robodoo r+

PR #395, opened 2025-01-28
[IMP] handyman:  Use costs/margins as materials budget 

dhrs-odoo:
In this PR we have 
- Remove "Materials Budget" from the SO line
- Rename "Material Budget" as "Allocated Budget"
- Added compute in materials budget
- Added Section Id field in sale order line

Task - 4506103


vava-odoo:
@robodoo r+

PR #396, opened 2025-01-29
[IMP] condominium: added a boolean field on analytic plan

dhrs-odoo:
In this PR we have added a boolean field on analytic plan in order to avoid direct using the reference of the xml id .

Task-4521719


vava-odoo:
See https://github.com/odoo/industry/pull/387

PR #409, opened 2025-02-03
[FIX] *: standard uom cannot be deleted

vava-odoo:


vava-odoo:
@robodoo r-

vava-odoo:
robodoo r+

PR #421, opened 2025-02-07
[FIX] condominium: different views for properties and analytic accounts

vava-odoo:
Before this commit, the view for properties (using specific analytic account) was shared with other analytic accounts. This prevents creating new analytic accounts that are not properties.
This commit solves the issue by using a new primary form view, linked to all window actions for properties.
It also removes window actions that were not used anymore.

task-4535537


vava-odoo:
@robodoo r+

PR #429, opened 2025-02-10
[FIX] industry_real_estate: Keep properties domain on refresh.

fame-odoo:
The properties domain was lost when refreshing the page, causing the view to display incorrect records.

This commit explicitly adds a `domain` field to ensure that properties are correctly filtered based on `plan_id`, maintaining consistency when the page is refreshed.

[opw-4417368](https://www.odoo.com/odoo/my-tasks/4417368)
Related commit: https://github.com/odoo/industry/commit/1718eaab37690cae2c1e9fc73f5065e20c562b7e

### DIFF
--- a/industry_real_estate/data/ir_actions_act_window.xml
+++ b/industry_real_estate/data/ir_actions_act_window.xml
@@ -31,6 +31,7 @@
         <field name="name">Properties</field>
         <field name="res_model">account.analytic.account</field>
         <field name="view_mode">kanban,list,form</field>
+        <field name="domain">[('x_is_property', '=', True)]</field>
         <field name="view_ids" eval="[
             (5, 0, 0),
             (0, 0, {'view_mode': 'kanban', 'view_id': ref('property_kanban_view')}),
--- a/industry_real_estate/data/ir_actions_server.xml
+++ b/industry_real_estate/data/ir_actions_server.xml
@@ -6,10 +6,8 @@
         <field name="state">code</field>
         <field name="code">
 <![CDATA[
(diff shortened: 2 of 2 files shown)


vava-odoo:
Hi @fame-odoo. Thanks for the fix!
I thought that 17.0 was not affected by this bug, and fixed it in saas~17.2 in https://github.com/odoo/industry/pull/353
Could you backport the whole diff from the PR? I think the server action can be cleaned as well.
Thanks!

vava-odoo:
Actually, there is kind of the same issue for the context. After a refresh, the context from the server action is lost and creating a new property is impossible (because there is no default account.plan anymore and the field is hidden). Could you fix that one as well? Thanks!

fame-odoo:
> Hi @fame-odoo. Thanks for the fix! I thought that 17.0 was not affected by this bug, and fixed it in saas~17.2 in #353 Could you backport the whole diff from the PR? I think the server action can be cleaned as well. Thanks!

Backport with the new changes you made is done.

> Actually, there is kind of the same issue for the context. After a refresh, the context from the server action is lost and creating a new property is impossible (because there is no default account.plan anymore and the field is hidden). Could you fix that one as well? Thanks!

I am working on this now, thanks!

vava-odoo (approved):
@robodoo r+ fw=no

Thanks!
For the other bug, I created [another task](https://www.odoo.com/odoo/all-tasks/4570800). If you already found a solution or want to continue working on this, don't hesitate to create another PR.
Cheers!

PR #435, opened 2025-02-14
[IMP] tests: add check for pot update

vava-odoo:
When updating a module in stable, the translation pot file should sometimes be updated as well for terms to be sent to transifex and translated. As it is not automated, this can be forgotten in the commit.

This commit therefore adds a test that verifies that the pot contains all translatable terms.


vava-odoo:
@robodoo r+

PR #477, opened 2025-02-26
[ADD] shoe_maker: add industry for Shoe Maker Services

vava-odoo:
This industry package includes all the essential features to manage your shoe maker services business, from customer relationship management to task and stock management as well as billing.

Basics
- Leverage the Sales App to manage your quotations and orders.
- Follow up on work completion with the Project App and Field Service App.
- Easily bill customers with the Point of Sales App.
- Slim your labor for stock management, replenishment and purchase with Inventory App and Purchase App.
- Automatically notify your customers when their order are ready to collect.

Included Customization
- On SO turning to collectible, the customer is notified by email to collect the order.

task-4529579


vava-odoo:
@robodoo r+

PR #486, opened 2025-03-04
[FIX] industry_restaurant: remove useless appointment overrides

vava-odoo:
Before saas-18.2, booking a table was displaying a default text. An override in this industry was used to modify the text. Now that the standard modules improved the text, this is no longer useful.


vava-odoo:
@robodoo r+

vava-odoo:
@robodoo r+

PR #493, opened 2025-03-07
[FIX] beverage_distributor,micro_brewery: remove useless filter_domain

vava-odoo:
`filter_domain` has no effect on `on_change` triggered automation rules. This behaviour will change in saas-18.3.`


vava-odoo:
@robodoo r+

PR #496, opened 2025-03-07
[FIX] beverage_distributor,micro_brewery: adapt filter_domain

vava-odoo:
Before this commit*, `filter_domain` had no effect on `on_change` triggered automation rules. Afterwards, it is taken into account and this `filter_domain` needs to be adapted.

*https://github.com/odoo/odoo/commit/64d6b25aa763a27c74eb479210a1313131c654e9


vava-odoo:
@robodoo r+

vava-odoo:
@robodoo r+

PR #513, opened 2025-03-17
[UPG] *: remove partner_gid and additional_info on res.partner

vava-odoo:
Since this commit*, these two fields are not defined on res.partner anymore. This commit removes their definition.

*https://github.com/odoo/odoo/commit/7814c1915d7c11ce9858dd5889e9871628f52096


vava-odoo:
@robodoo r+

PR #527, opened 2025-03-21
[IMP] tests: improve test on pot changes

vava-odoo:
Before this commit, the test on pot change was not robust, and the displayed diff was not accurate.

With this commit, the first lines of the pot are omitted (not important for translations purposes), the diff is better constructed (row by row) and the displayed diff is cleaner.


vava-odoo:
@robodoo r+

PR #533, opened 2025-03-25
[FIX] marketing_agency: do not send automatic emails

vava-odoo:


vava-odoo:
@robodoo help

vava-odoo:
@robodoo r+ fw=skipci

PR #541, opened 2025-03-27
[FIX] industry_lawyer: remove wrong downpayment SO lines

vava-odoo:


vava-odoo:
@robodoo fw=no

vava-odoo:
@robodoo r+

PR #549, opened 2025-04-03
[FIX] fitness: disabled restrict categories in pos configuration

dhrs-odoo:
In this PR, we have disabled the Restrict Categories setting.

Task-4700036

### DIFF
--- a/fitness/data/pos_config.xml
+++ b/fitness/data/pos_config.xml
@@ -6,7 +6,6 @@
         <field name="iface_print_auto" eval="True"/>
         <field name="ship_later" eval="True"/>
         <field name="customer_display_type">local</field>
-        <field name="limit_categories">True</field>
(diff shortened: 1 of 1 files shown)


fitness/data/res_config_settings.xml
```diff
@@ -5,6 +5,7 @@
         <field name="group_discount_per_so_line" eval="1"/>
         <field name="group_product_pricelist" eval="True"/>
         <field name="pos_use_pricelist" eval="True"/>
+        <field name="pos_limit_categories" eval="False"/>
```
vava-odoo, line 8: Shouldn't this setting be removed from the pos config instead?

dhrs-odoo: oops :fearful:  you are right

vava-odoo (approved):
@robodoo r+

PR #557, opened 2025-04-08
[UPG] *: ir.filter field user_id became user_ids

vava-odoo:
This commit* changes the user_id (m2o) field to user_ids (m2m) field. In industry occurences, the filter is shared so there is no need to define the user anyway, so the field definition is removed.

*https://github.com/odoo/odoo/commit/414e55cf7c3971a7ba0a7c96855db4c23e77a5c4


vava-odoo:
@robodoo r+

PR #574, opened 2025-04-16
[FIX] industry_real_estate: fix navigation on website

dhrs-odoo:
In this PR we have fixed broken navigation for the **"About Us"** and **"Properties"** menu 
items on the website.

Task-4700144


vava-odoo (approved):
@robodoo r+

PR #577, opened 2025-04-17
[FIX] automobile: remove search from warranty rule to avoid loop

jobl-odoo:
### Issue:
The `automobile` industry module contains an Automation Rule to set the `x_warranty_date` value, a stored field on `stock.move.line` related to `stock.lot.expiration_date`. The rule has a filter to ensure that the update is only performed if the record meets those criteria. However, we then `search` for all other records that meet the criteria, and update each found record based on the current time. This causes 2 issues.

1) Each record is getting its `x_warranty_date` updated based on the time that one specific record is saved.
2) More pressing, the `search` followed by a `write` creates a loop that causes the Automation Rule to trigger itself infinitely, never allowing the save to complete.

### Solution:
We can remove the `search` and apply the rule to the specific record that was updated, solving both problems simultaneously.

opw-4672112

### DIFF
--- a/automobile/data/ir_actions_server.xml
+++ b/automobile/data/ir_actions_server.xml
@@ -6,9 +6,8 @@
         <field name="state">code</field>
         <field name="base_automation_id" ref="update_the_warranty"/>
         <field name="code"><![CDATA[
-move_lines = records.search([("state", "=", "done"), ("picking_code", "=", "outgoing")])
-for line in move_lines:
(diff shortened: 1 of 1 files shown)


vava-odoo (approved):
Thanks @jobl-odoo for the fix.
Indeed, this is a real mess :grimacing: 
@robodoo r+ squash
(squash to edit the opw number)
Note that you'll need to correct this on the db (manually or upgrade the module).

PR #578, opened 2025-04-18
[UPG] micro_brewery: remove notes on mrp.routing.workcenter

vava-odoo:
A commit removes the detailed notes on mrp.routing.workcenter. This commit therefore removes it. We'll improve the data in a subsequent task.


vava-odoo:
robodoo r+

PR #590, opened 2025-04-18
[ADD] base_industry_data: add base module for industry

vava-odoo:
This commit adds a base module for all industry, with some shared records: partners, tour, ...
This will reduce duplicated entries and paves the way to have shared sub-modules between industries

task-4575230


vava-odoo:
@robodoo r+

PR #630, opened 2025-05-12
[IMP] base_industry_module,*: move records to base module

vava-odoo:
Since this previous commit*, there is a base data module, and marketing_agency module was the first test. Since the test is working as expected, this commit goes further and move every common records to this base module.

task-4575230

*https://github.com/odoo/industry/commit/8f4c1ef24f6b5697f586c2decf577484a5db2959


vava-odoo:
@robodoo r+

PR #647, opened 2025-05-15
[IMP] industry_restaurant:upgrade

frva-odoo:
This commit adds new features for the industry_restaurant module such as  kiosk attendance, POS rights for employees, pricelists presets (Happy hour) and tables booking.
It alse reworks demo data by adding employees, changing some product's categories, changing the working schedules and adding a demo takeout order.

TASK-4489370


industry_restaurant/__manifest__.py
```diff
@@ -1,11 +1,12 @@
 {
     'name': 'Fine Dining Restaurant',
-    'version': '1.0',
+    'version': '3.0',
```
vava-odoo:
```suggestion
    'version': '1.1',
```

industry_restaurant/__manifest__.py
```diff
@@ -1,11 +1,12 @@
...
     'category': 'Hospitality',
     'depends': [
         'account_followup',
         'contacts',
         'hr',
```
vava-odoo:
```suggestion
```

industry_restaurant/data/ir_default_data.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="1">
```
vava-odoo: data is another level, useful when dealing with updatable and non-updatable records. Here, there is only one type, so only odoo is fine
```suggestion
<odoo noupdate="1">
```

industry_restaurant/data/product_pricelist.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+  <record id="product_pricelist_1" model="product.pricelist">
+    <field name="name">Default</field>
+  </record>
```
vava-odoo, line 5: Default pricelist is defined twice... I think you need to define it before activating the config setting to avoid the other default one to be created

industry_restaurant/data/resource_calendar_data.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo noupdate="1">
+    <record id="resource.resource_calendar_std" model="resource.calendar">
```
vava-odoo: What happens if this record was deleted before? Without anything, it'll raise an error. You need to add `forcecreate`

Also, from the spec, I think this should be in demo, right?

industry_restaurant/demo/appointment_type.xml
```diff
@@ -2,6 +2,14 @@
...
     <record id="appointment_type_1" model="appointment.type">
         <field name="is_published" eval="True"/>
-        <field name="resource_ids" model="appointment.resource" eval="[(6, 0, obj().search(['|', ('name', 'ilike', 'Patio'), ('name', 'ilike', 'Floor')]).ids)]"/>
+        <field name="image_1920" type="base64" file="industry_restaurant/static/src/binary/appointment_resource/table1.jpg"/>
+        <field name="location_id" ref="base.main_partner"/>
+        <field name="assign_method">time_auto_assign</field>
+        <field name="category">recurring</field>
+        <field name="avatars_display">hide</field>
+        <field name="max_schedule_days">30</field>
+        <field name="slot_creation_interval">0.25</field>
```
vava-odoo: Why do you need to re-define all these fields? Some are already defined in data/, no need to define the exact same thing here (example: `<field name="assign_method">time_auto_assign</field>`)

industry_restaurant/demo/hr_employee.xml
```diff
@@ -1,24 +1,64 @@
...
         <field name="address_id" ref="base.main_partner"/>
         <field name="work_phone">+1 650-691-3277</field>
-        <field name="name">Chef Jeff</field>
+        <field name="work_email">casey.owner@yourcompany.example.com</field>
+        <field name="pin">3333</field>
     </record>
-    <record id="hr_employee_2" model="hr.employee">
-        <field name="work_contact_id" ref="res_partner_15"/>
-        <field name="job_title">Service</field>
+    <record id="hr_employee_3" model="hr.employee" forcecreate="1">
```
vava-odoo: forcecreate only makes sense on records of other modules
```suggestion
    <record id="hr_employee_3" model="hr.employee">
```

industry_restaurant/demo/pos_config.xml
```diff
@@ -1,4 +1,17 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo noupdate="1">
-    <function model="pos.config" name="load_onboarding_restaurant_scenario" />
+    <function model="pos.config" name="load_onboarding_restaurant_scenario"/>
+    <record id="pos_restaurant.pos_config_main_restaurant" model="pos.config" forcecreate="0">
+        <field name="module_pos_hr" eval="True"/>
```
vava-odoo, line 5: To me, `module_*` settings should be in data/, no?

industry_restaurant/demo/pos_order_lines.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="pos_order_line_1" model="pos.order.line">
+        <field name="product_id" ref="pos_restaurant.pos_food_cheeseburger"/>
+        <field name="order_id" ref="pos_order_1"/>
+        <field name="price_unit">13.0</field>
+        <field name="price_subtotal_incl">14.95</field>
+        <field name="price_subtotal">13.0</field>
+        <field name="tax_ids" eval="[ref('account.1_sale_tax_template')]"/>
```
vava-odoo: I doubt we need to add it here (to me this is automated)
```suggestion
```

industry_restaurant/demo/product_product.xml
```diff
@@ -20,4 +20,31 @@
             <field name="route_ids" eval="[(6, 0, [ref('purchase_stock.route_warehouse0_buy')])]"/>
             <field name="seller_ids" eval="[(6, 0, [ref('product_supplierinfo_6')])]"/>
         </record>
+        <record id="pos_restaurant.coke" model="product.product">
```
vava-odoo: forcecreate

industry_restaurant/static/description/index.html
```diff
@@ -1,6 +1,21 @@
...
+  <li> The Kitchen Display will ensure a great follow-up of every order in your bar and kitchen.</li>
+  <li> Let your customer book a table anytime with the Appointment App.</li>
+  <li> Use the Project App to never miss a reordering or a cleaning task.</li>
+  <li> Use the Planning App to schedule and share your shifts with your employees.</li>
+  <li> Use the Purchase App to reorder your products.</li>
+  <li> Use the Inventory App to manage your stock and receive products.</li>
+  <li> Get visibility and share your menu with a great Website.</li>
+  <li> Use the Attendances App to record employee's time spent working.</li>
+</ul>
+Customisations
```
vava-odoo:
```suggestion
Customizations
```

vava-odoo (commented):
Good! Just a few comments/questions/remarks reading through. Let me know if anything is unclear.

industry_restaurant/demo/pos_order_lines.xml
frva-odoo: I added it because it is not showing if I don't, I checked and it's needed to be displayed on the line

industry_restaurant/demo/pos_config.xml
frva-odoo: As I understand it, this is a setting available for a specific pos, and here I create the restaurant pos in demo so I can't put it in data, unless I also move the pos restaurant creation (which is the restaurant scenario with demo data), am I right ?

vava-odoo: Ah, yes, it is in pos.config, not res.config.settings :facepalm:

industry_restaurant/demo/pos_order_lines.xml
vava-odoo: Ok then. Could you take caution and add `raise_if_not_found=False`? Thanks

industry_restaurant/demo/appointment_type.xml
```diff
@@ -2,6 +2,11 @@
 <odoo noupdate="1">
     <record id="appointment_type_1" model="appointment.type">
         <field name="is_published" eval="True"/>
-        <field name="resource_ids" model="appointment.resource" eval="[(6, 0, obj().search(['|', ('name', 'ilike', 'Patio'), ('name', 'ilike', 'Floor')]).ids)]"/>
+        <field name="location_id" ref="base.main_partner"/>
+        <field name="avatars_display">hide</field>
+        <field name="max_schedule_days">30</field>
+        <field name="slot_creation_interval">0.25</field>
+        <field name="staff_user_ids" eval="[(6, 0, [])]"/>
+        <field name="resource_ids" model="appointment.resource" eval="[(6, 0, obj().search([('capacity','>','1'),('name', 'ilike', 'Table%')]).ids)]"/>
```
vava-odoo: Isn't `capacity` an integer?
```suggestion
        <field name="resource_ids" model="appointment.resource" eval="[(6, 0, obj().search([('capacity', '>', 1),('name', 'ilike', 'Table%')]).ids)]"/>
```

vava-odoo (approved):
Thanks for this!

vava-odoo (approved):
@robodoo r+

PR #661, opened 2025-05-22
[FIX] industry_real_estate: improve performance on computed x_rental_contract_id

frva-odoo:
Before this commit, x_rental_start_date was defined after being used in the computed field x_rental_contract_id, which led to sql_db deadlock. Furthermore, the performance of the computed method was poor due to a search on sale.order in a loop.
This commit moves the definition of x_rental_contract_id after its dependency fields definitions, and improves the performance of the computed field by using _read_group method instead.
Additionally, it adds a new test to verify that x_rental_contract_id is correctly computed.

OPW-4805375

### DIFF
--- a/industry_real_estate/data/ir_model_fields.xml
+++ b/industry_real_estate/data/ir_model_fields.xml
@@ -184,22 +184,6 @@ for property in self:
         <field name="ttype">many2many</field>
     </record>
 
-    <record id="field_property_rental_contract_id" model="ir.model.fields">
-        <field name="name">x_rental_contract_id</field>
-        <field name="field_description">Rental Contract</field>
-        <field name="model_id" ref="analytic.model_account_analytic_account"/>
-        <field name="relation">sale.order</field>
-        <field name="ttype">many2one</field>
-        <field name="readonly" eval="True"/>
-        <field name="store" eval="False"/>
-        <field name="compute">
-<![CDATA[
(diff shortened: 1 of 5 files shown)


industry_real_estate/data/ir_model_fields.xml
```diff
@@ -239,6 +223,24 @@ for record in self:
...
+        <field name="name">x_rental_contract_id</field>
+        <field name="field_description">Rental Contract</field>
+        <field name="model_id" ref="analytic.model_account_analytic_account"/>
+        <field name="relation">sale.order</field>
+        <field name="ttype">many2one</field>
+        <field name="readonly" eval="True"/>
+        <field name="store" eval="False"/>
+        <field name="compute">
+<![CDATA[
+sos = self.env['sale.order']._read_group([('x_account_analytic_account_id', 'in', self.ids), ('state', '=', 'sale'), ('x_rental_start_date', '<=', datetime.date.today())], groupby=['x_account_analytic_account_id'], aggregates=['x_rental_start_date:max','id:min'])
```
vava-odoo, line 237: Looking twice, this fails when there are several SOs linked to the same AA.

vava-odoo (commented):
Just a few remarks/comments/suggestions on the test, otherwise it's good to go. Thanks @frva-odoo

PR #671, opened 2025-05-30
"[FIX] tests: removing price_unit from the USELESS_FIELDS"

frva-odoo:
Removing the pos.order.line field "price_unit" from the USELESS_FIELDS dictionary in the tests because it is needed to change the unit price on an order line.


vava-odoo (approved):
@robodoo r+

PR #679, opened 2025-06-05
[FIX] industry_lawyer: fix wrong title in app description

arthurrrm:
The title was referencing the wrong app (Handyman Services), it now shows Law Firm instead.

task-4848479


vava-odoo (approved):
Thanks! Fix is correct.
Could you just change `industry_lawfirm` by `industry_lawyer` (that's the technical name of the module)?

PR #680, opened 2025-06-05
[FIX] *: remove timesheets emails in demo for modules depending ts

arthurrrm:
We prevent theses emails by adding to each of these modules a record of res.config.settings with the fields "reminder_user_allow" and "reminder_allow" set to False (inherited from timesheet_grid)

task-4653309


food_trucks/__manifest__.py
arthurrrm, line 1: (whole file had wrong identation of only 2 spaces)

PR #686, opened 2025-06-06
[ADD] fitness: introduce the use of Membership products

arthurrrm:
before this commit, server actions and automations where used to update the pricelist of a partner. Now, with the use of the new partnership/membership module, this is automated and users can link a grade of membership (E.g. Silver, Gold...) to a product. This will automatically set/unset the pricelist (associated with this grade) of the partner when the membership start/is revoked

task-4853044


fitness/data/res_partner_grade.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo noupdate="1">
+    <record id="partnership.res_partner_grade_data_bronze" model="res.partner.grade">
```
vava-odoo: if you edit a record from another module, you need to define the behaviour in case the record was removed. In this case, we don't want to create it back if it was deleted, so 
```suggestion
    <record id="partnership.res_partner_grade_data_bronze" model="res.partner.grade" forcecreate="0">
```

vava-odoo (approved):

fitness/__manifest__.py
vava-odoo, line 1: Forgot this point, but could you bump the version to 1.1? Thanks

vava-odoo (commented):
@robodoo r+

PR #688, opened 2025-06-10
[FIX] tests: test noupdate flag for functions

arthurrrm:
The `noupdate` flag is also checked for functions, before this commit it was only checked for records.
Additionally, the tests now support the nuopdate flag in \<data> tags, surrounding the records/functions

### DIFF
--- a/condominium/data/ir_actions_server.xml
+++ b/condominium/data/ir_actions_server.xml
@@ -5,43 +5,44 @@
         <field name="state">code</field>
         <field name="model_id" ref="sale.model_sale_order"/>
         <field name="code"><![CDATA[
-distributions = record.sudo().pricelist_id.x_ratios # By default, we assume that there is a pricelist/distribution key set on the SO
-owner_field = 'x_owner'
-key_field = 'x_ratio'
-analytic_account_field = 'x_property'
-total_distribution = sum(distributions.mapped(key_field))
-if total_distribution == 0: # If there is no pricelist/distribution key set on the SO, we use the area as distribution
-    distributions = record.sudo().partner_id.x_condominium_properties
-    owner_field = 'partner_id'
-    key_field = 'x_area'
-    analytic_account_field = '' # There is no field since the property is an analytic account
+if record.state == 'draft':
+    distributions = record.sudo().pricelist_id.x_ratios # By default, we assume that there is a pricelist/distribution key set on the SO
+    owner_field = 'x_owner'
+    key_field = 'x_ratio'
+    analytic_account_field = 'x_property'
     total_distribution = sum(distributions.mapped(key_field))
-if total_distribution == 0:
-    raise UserError("Cannot split the Sales Orders because there is not distribution key in the pricelist nor areas")
-owners = env['res.partner'].search([('x_companies', 'in', record.company_id.id)])
-for owner in owners:  # One SO per owner
-    new_so = record.copy()
-    new_so['x_source_sales_order'] = record.id
-    new_so['partner_id'] = owner.id
-    new_so.order_line.unlink() # Remove the lines that were copied from the source SO
-    new_so_lines = []
-    for sol in record.order_line: # The lines are re-created using the distribution key and the analytic account corresponding to the property is set
-        for distribution in distributions:
-            if distribution[owner_field].id == owner.id:
-                analytic_account = distribution if not analytic_account_field else distribution[analytic_account_field]
-                new_so_lines.append((0, 0, {
-                    'order_id': record.id,
-                    'product_id': sol.product_id.id,
-                    'name': sol.name,
-                    'product_uom_qty': sol.product_uom_qty,
-                    'product_uom_id': sol.product_uom_id.id,
-                    'price_unit': sol.price_unit * distribution[key_field] / total_distribution,
-                    'tax_ids': sol.tax_ids,
-                    'discount': sol.discount,
-                    'analytic_distribution': {analytic_account.id: 100},
-                }))
-    new_so.write({'order_line': new_so_lines})
-record['state'] = 'cancel'
+    if total_distribution == 0: # If there is no pricelist/distribution key set on the SO, we use the area as distribution
+        distributions = record.sudo().partner_id.x_condominium_properties
+        owner_field = 'partner_id'
+        key_field = 'x_area'
+        analytic_account_field = '' # There is no field since the property is an analytic account
+        total_distribution = sum(distributions.mapped(key_field))
+    if total_distribution == 0:
+        raise UserError("Cannot split the Sales Orders because there is not distribution key in the pricelist nor areas")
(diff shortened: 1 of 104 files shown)


tests/test_generic/tests/test_xml.py
```diff
@@ -67,6 +67,7 @@
     "ir.ui.view",
     "knowledge.article",
     "website.controller.page",
+    "web_editor.assets",
```
vava-odoo, line 72: From the discussion I just had with the PO, you can add `ir.module.module` to this list as well.

tests/test_generic/tests/test_xml.py
```diff
@@ -265,18 +266,25 @@ def _check_xml_style(self, s, root, module, file_name):
...
-        for record in root.xpath("//record"):
+        for record in root.xpath("//record") + root.xpath("//function"):
             model = record.get('model')
+            noupdate = False
+            parent = record.getparent()
+            while parent is not None:  # Find nearest parent with 'noupdate' attribute (data tag or odoo header tag)
+                if 'noupdate' in parent.attrib:
+                    noupdate = parent.attrib['noupdate'] in ('1', 'True', 'true')
+                    break
+                parent = parent.getparent()
```
vava-odoo: I'm not sure whether this is true, but its definitely not a good practice. Usually, the `noupdate` is only defined at one (sub-)level. The record/function will get the one from his (closest) parent. If undefined, the default value is `False` (`noupdate="0"`). The parent of the parent shouldn't have any impact.

Such a case shouldn't exist, but is the record update or noupdate?
```xml
<odoo noupdate="1">
  <data>
    <record ...>
      ...
    </record>
  </data>
</odoo>
```

If the answer is noupdate, I wonder if this shouldn't be another test: if there is several levels, don't set the noupdate=1 to the outer level. What do you think?

vava-odoo (commented):
Good! it solves the issue for sales and purchase orders. 
About website, the idea is to get the same website after a reload, so I think we need to change a bit.
Let me know if you don't agree with this.
Thanks for your work!

condominium/demo/sale_order_post.xml
```diff
@@ -1,18 +1,24 @@
...
+        <function model="ir.model.data" name="_update_xmlids">
+            <value model="base" eval="[{
+                'xml_id': 'condominium.product_pricelist_1',
+                'record': obj().env.ref('condominium.sale_order_154').pricelist_id,
+                'noupdate': True,
+            }]"/>
+        </function>
+    </data>
+    
     <function name="run" model="ir.actions.server" context="{'active_model': 'product.pricelist', 'active_ids': ref('product_pricelist_1')}">
```
vava-odoo, line 13: This one is an issue... Running this twice will generate mistakes in the db. I need to check the result, and see if we can live with it

marketing_agency/demo/website_theme_apply.xml
```diff
@@ -6,8 +6,10 @@
         <value model="ir.module.module" eval="obj().env['ir.module.module'].search([('name', '=', 'theme_buzzy')]).ids"/>
         <value model="ir.module.module" eval="obj().env.ref('website.default_website')"/>
     </function>
-    <function model="theme.utils" name="enable_view" eval="['website.template_header_hamburger']"/>
-    <function model="theme.utils" name="enable_view" eval="['website.no_autohide_menu']"/>
+    <data noupdate="1">
+        <function model="theme.utils" name="enable_view" eval="['website.template_header_hamburger']"/>
```
vava-odoo: Add this model as well, forgot about this one

micro_brewery/demo/website_theme_apply.xml
```diff
@@ -1,26 +1,30 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo>
+
```
vava-odoo: useless diff

non_profit_organization/demo/website_theme_apply.xml
```diff
@@ -1,11 +1,13 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo>
+
```
vava-odoo: same

toy_store/demo/website_theme_apply.xml
```diff
@@ -6,7 +6,6 @@
         <value model="ir.module.module" eval="obj().env['ir.module.module'].search([('name', '=', 'theme_kiddo')]).ids"/>
         <value model="ir.module.module" eval="obj().env.ref('website.default_website')"/>
     </function>
-
```
vava-odoo: same

tests/test_generic/tests/test_xml.py
```diff
@@ -265,18 +267,32 @@ def _check_xml_style(self, s, root, module, file_name):
...
+        for record in root.xpath("//record") + root.xpath("//function"):
             model = record.get('model')
+            noupdate = False
+            parent = record.getparent()
+            data_tag = False
+            while parent is not None:  # Find nearest parent with 'noupdate' attribute (data tag or odoo header tag)
+                if 'noupdate' in parent.attrib:
+                    if data_tag:
+                        _logger.warning(
+                            "Cannot set 'noupdate' arround an already existing 'data' tag in %s",
```
vava-odoo:
```suggestion
                            "Avoid setting 'noupdate' around an already existing 'data' tag in %s",
```

tests/test_generic/tests/test_xml.py
```diff
@@ -265,18 +267,32 @@ def _check_xml_style(self, s, root, module, file_name):
...
+                            filename,
+                        )
+                    noupdate = parent.attrib['noupdate'] in ('1', 'True', 'true')
```
vava-odoo: Don't you need to leave the while as soon as noupdate is true?

condominium/data/ir_actions_server.xml
```diff
@@ -5,43 +5,44 @@
...
-owner_field = 'x_owner'
-key_field = 'x_ratio'
-analytic_account_field = 'x_property'
-total_distribution = sum(distributions.mapped(key_field))
-if total_distribution == 0: # If there is no pricelist/distribution key set on the SO, we use the area as distribution
-    distributions = record.sudo().partner_id.x_condominium_properties
-    owner_field = 'partner_id'
-    key_field = 'x_area'
-    analytic_account_field = '' # There is no field since the property is an analytic account
+if record.state == 'draft':
```
vava-odoo, line 8: Couldn't you reduce the diff with an early return?
```suggestion
if record.state != 'draft': return
```

vava-odoo (approved):
Thanks!

condominium/data/ir_actions_server.xml
vava-odoo: nope

vava-odoo:
@robodoo r+

PR #689, opened 2025-06-12
[ADD] utils: Util to launch an industry easily

arthurrrm:
```
Usage: ./industry/run_industry.sh -i <industry-name> [-d] [-t] [-r]
  -i <industry-name>   (re)Install this industry
  -d                   Enable demo data when installing
  -t                   Run tests for the installed industry
  -r                   Reset the database before running
```

### DIFF
--- a/run_industry.sh
+++ b/run_industry.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Usage function
+usage() {
+  echo "Usage: ./industry/run_industry.sh -i <industry-name> [-d] [-t] [-r]"
+  echo "  -i <industry-name>   (re)Install this industry"
+  echo "  -d                   Enable demo data when installing"
+  echo "  -t                   Run tests for the installed industry"
+  echo "  -r                   Reset the database before running"
+  exit 1
+}
+
+# Default values
+INDUSTRY_NAME=""
+INSTALL=false
+TEST=false  
+RESET_DB=false
+DEMO=False
+
+# Parse arguments
+while getopts ":i:dtr" opt; do
+  case $opt in
+    i) INDUSTRY_NAME="$OPTARG"; INSTALL=true ;;
+    d) DEMO=True ;;
+    t) TEST=true ;;
+    r) RESET_DB=true ;;
+    *) usage ;;
+  esac
+done
+
+
+echo "Industry: $INDUSTRY_NAME"
+echo "Install: $INSTALL"
+echo "Demo: $DEMO"
+echo "Test: $TEST"
+echo "Reset DB: $RESET_DB"
+
+PYTHON_BIN="python3"
+ODOO_BIN="odoo/odoo-bin"
+ADDONS_PATH="industry/tests,enterprise,odoo/addons,odoo/odoo/addons,design-themes"
+TEST_TAGS="/test_generic,/test_$INDUSTRY_NAME"
+DB="test"
+
+if $RESET_DB; then
+  echo "Resetting database '$DB'..."
+  dropdb --if-exists "$DB"
+  $PYTHON_BIN $ODOO_BIN --addons-path="$ADDONS_PATH" -i base_import_module,test_generic,test_$INDUSTRY_NAME -d $DB --without-demo=1 --stop-after-init
+fi
+
+if $INSTALL; then
+  echo "Initializing modules..."
+
+  TMP_INSTALL_PY=$(mktemp)
+  cat <<EOF > "$TMP_INSTALL_PY"
+import sys
+sys.path.append('industry/')
+from utils import get_zip
+def main():
+    zip = get_zip('$INDUSTRY_NAME', env)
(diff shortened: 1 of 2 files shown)


zip_industry.py
vava-odoo, line 1: Could you rename it utils.py?

zip_industry.py
```diff
@@ -0,0 +1,31 @@
...
+import os
+from io import BytesIO
+from zipfile import ZipFile
+
+
+def get_zip(module_name, env):
+    output = BytesIO()
+    zf = ZipFile(output, "w")
+    external_modules = get_external_dependencies('industry/' + module_name, env)
+    dirs = ['industry/' + module_name] + ['industry/' + module for module in external_modules]
```
vava-odoo: Could you check the runbot code as well? I think the path is different, and we'd need to find an unified way

zip_industry.py
```diff
@@ -0,0 +1,31 @@
...
+    for dir in dirs:
+        for dirname, subdirs, files in os.walk(dir):
+            zip_dirname = os.sep.join(dirname.split(os.sep)[1:])
+            zf.write(dirname, zip_dirname)
+            for filename in files:
+                zf.write(os.path.join(dirname, filename), os.path.join(zip_dirname, filename))
+    zf.close()
+    output.seek(0)
```
vava-odoo: Why do you need this?

zip_industry.py
```diff
@@ -0,0 +1,31 @@
...
+    return output
+
+def get_external_dependencies(dir, env):
+    manifest_file = dir + '/__manifest__.py'
+    if not os.path.exists(manifest_file):
+        print(f"Manifest not found: {manifest_file}")
+        return set()
```
vava-odoo: If there is no manifest, nothing will be installed anyway. I wonder if this shouldn't raise instead of failing silently... Other solution is to replace the print by a warning in the log

zip_industry.py
```diff
@@ -0,0 +1,31 @@
...
+    with open(manifest_file, mode='r') as f:
+        manifest = ast.literal_eval(f.read())
```
vava-odoo: Maybe more secure to use file_open(..., env=env)

run_industry.sh
```diff
@@ -0,0 +1,82 @@
...
+
+  TMP_INSTALL_PY=$(mktemp)
+  cat <<EOF > "$TMP_INSTALL_PY"
+import sys
+
+sys.path.append('industry/')
+from zip_industry import get_zip
+def main():
+    zip = get_zip('$INDUSTRY_NAME', env)
+    res = env['ir.module.module'].sudo()._import_zipfile(zip, force=False, with_demo=True)
```
vava-odoo: Would be nice to be able to select if we add demo data or not + I'd force a fresh install 
```suggestion
    res = env['ir.module.module'].sudo()._import_zipfile(zip, force=True, with_demo=with_demo)
```

run_industry.sh
```diff
@@ -0,0 +1,82 @@
...
+
+if $INSTALL; then
+  echo "Initializing modules..."
+  $PYTHON_BIN $ODOO_BIN --addons-path="$ADDONS_PATH" -i base_import_module,test_generic,test_$INDUSTRY_NAME -d $DB --without-demo=1 --stop-after-init
+
+  TMP_INSTALL_PY=$(mktemp)
+  cat <<EOF > "$TMP_INSTALL_PY"
+import sys
+
+sys.path.append('industry/')
```
vava-odoo, line 57: I don't get why you need this

vava-odoo:
You'll also have to edit the test so that these new files are not considered as industry modules

frva-odoo:
> @frva-odoo would you mind having a look at this?

I have nothing to add after reading your comments, everything looks good to me 🙂

zip_industry.py
arthurrrm: problem is that industry/ is not in the addons path so we cant use this

```
def file_open(name: str, mode: str = "r", filter_ext: tuple[str, ...] = (), env: Environment | None = None):
    """Open a file from within the addons_path directories, as an absolute or relative path.
```

utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+def get_external_dependencies(dir, env):
+    manifest_file = dir + '/__manifest__.py'
+    if not os.path.exists(manifest_file):
+        raise FileNotFoundError(f"Manifest file not found: {manifest_file}")
+    with file_open_temporary_directory(env) as temp_dir:
+        temp_manifest_path = os.path.join(temp_dir, '__manifest__.py')
+        shutil.copy(manifest_file, temp_manifest_path)
+        with file_open(temp_manifest_path, env=env) as f:
+            manifest = ast.literal_eval(f.read())
```
vava-odoo: That's not exactly how I wanted the file_open to behave :sweat_smile: 
No meaning if we're to copy everything there.
Sorry, wrong way

utils.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
```
vava-odoo: not sure we need this

vava-odoo (approved):
@robodoo r+

PR #691, opened 2025-06-13
[FIX] industry_real_estate: create a property after refreshing the page

arthurrrm:
This was caused by the loss of the server action used to set the default value for the account plan, we now use a ir.default record to set this.

task-4570800


industry_real_estate/data/ir_actions_server.xml
```diff
@@ -7,7 +7,6 @@
         <field name="code">
 <![CDATA[
 action = env['ir.actions.actions']._for_xml_id('industry_real_estate.action_properties')
```
vava-odoo: Is the server action still useful then? Can't we link the menu item to the act window directly?

vava-odoo (approved):
@robodoo r+
As discussed in the task, we need to check which version is impacted and adapt the forward port if needed

PR #696, opened 2025-06-17
[IMP] hvac_services: Add industry module for HVAC

frva-odoo:
Base implementation of the hvac_services module.
It provides a first set of services, products, worksheets and reporting to install, maintain and repair HVAC devices on your customer sites with detailed traceability.

TASK-4032364


hvac_services/__manifest__.py
```diff
@@ -0,0 +1,83 @@
+{
+    'name': 'HVAC Services',
+    'version': '1.0',
+    'category': 'Services',
+    'depends': [
+        'appointment_account_payment',
+        'contacts',
+        'hr',
+        'industry_fsm',
```
vava-odoo: Not needed, because already in the dependency tree via industry_fsm_sale (we try to reduce the list for clarity on Appstore
```suggestion
```

hvac_services/__manifest__.py
```diff
@@ -0,0 +1,83 @@
...
+        'industry_fsm_sale_report',
+        'industry_fsm_stock',
+        'knowledge',
+        'project_purchase',
+        'project_sale_subscription',
+        'purchase',
```
vava-odoo: Same
```suggestion
```

hvac_services/__manifest__.py
```diff
@@ -0,0 +1,83 @@
...
+        'sale_management',
+        'sale_pdf_quote_builder',
```
vava-odoo: Not used, and auto-installed
```suggestion
```

hvac_services/__manifest__.py
```diff
@@ -0,0 +1,83 @@
...
+        'sale_planning',
+        'sale_purchase_stock',
+        'sale_subscription',
+        'stock',
```
vava-odoo:
```suggestion
```

hvac_services/__manifest__.py
```diff
@@ -0,0 +1,83 @@
+{
+    'name': 'HVAC Services',
+    'version': '1.0',
+    'category': 'Services',
+    'depends': [
+        'appointment_account_payment',
+        'contacts',
```
vava-odoo, line 8: add base_industry_data, so that you don't have to define the tour, the partners, ...

hvac_services/data/ir_actions_act_window.xml
```diff
@@ -0,0 +1,21 @@
...
+    <record id="report_refrigerant_consumption_action" model="ir.actions.act_window">
+        <field name="name">Refrigerant Consumption</field>
+        <field name="domain">[('x_refrigerant_product','!=',False)]</field>
+        <field name="res_model">x_project_task_worksheet_template_2</field>
+        <field name="view_mode">pivot,form</field>
+    </record>
+    <record id="get_maintenances_action" model="ir.actions.act_window">
+        <field name="name">Maintenances</field>
+        <field name="context">{'search_default_x_maintained_serials': active_id,'default_x_maintained_serials': [active_id]}</field>
+        <field name="domain">[('x_maintained_serials', '=', active_id)]</field>
```
vava-odoo: If x_maintained_serials is a X2Many, then the convention to suffix it with _ids helps a lot. And I'm not sure it is a good practice to use '=' in this case

hvac_services/__manifest__.py
```diff
@@ -0,0 +1,83 @@
...
+        'project_purchase',
+        'project_sale_subscription',
+        'purchase',
+        'sale_management',
+        'sale_pdf_quote_builder',
+        'sale_planning',
+        'sale_purchase_stock',
+        'sale_subscription',
+        'stock',
+        'timesheet_grid',
```
vava-odoo: In industry_fsm
```suggestion
```

hvac_services/data/ir_actions_act_window_2.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo>
+    <record id="industry_fsm_report.x_project_task_worksheet_template_2_ir_actions_get_worksheets" model="ir.actions.act_window" forcecreate="1">
```
vava-odoo: I wonder why you need to edit an existing one... I'd rather create a new record for the current module instead

hvac_services/data/ir_actions_act_window_2.xml
```diff
@@ -0,0 +1,16 @@
...
+        <field name="name">Worksheets</field>
+        <field name="context">{'edit': False, 'create': False, 'delete': False, 'duplicate': False, 'action_xml_id': 'industry_fsm_report.fsm_worksheets_action_settings', 'worksheet_template_id': 2}</field>
```
vava-odoo: Sure about this? Especially the '2' seems louche

hvac_services/data/ir_actions_act_window_2.xml
```diff
@@ -0,0 +1,16 @@
...
+        <field name="res_model">x_project_task_worksheet_template_2</field>
+        <field name="search_view_id" ref="industry_fsm_report.x_project_task_worksheet_template_2_search_view"/>
+    </record>
+    <record id="display_stock_lot_list_view_with_maintenances" model="ir.actions.act_window">
+        <field name="name">Stock lot custom view</field>
+        <field name="res_model">stock.lot</field>
+        <field name="view_mode">list,form</field>
```
vava-odoo:
```suggestion
        <field name="view_mode">list</field>
```

hvac_services/data/ir_actions_act_window_2.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo>
+    <record id="industry_fsm_report.x_project_task_worksheet_template_2_ir_actions_get_worksheets" model="ir.actions.act_window" forcecreate="1">
+        <field name="name">Worksheets</field>
+        <field name="context">{'edit': False, 'create': False, 'delete': False, 'duplicate': False, 'action_xml_id': 'industry_fsm_report.fsm_worksheets_action_settings', 'worksheet_template_id': 2}</field>
+        <field name="res_model">x_project_task_worksheet_template_2</field>
+        <field name="search_view_id" ref="industry_fsm_report.x_project_task_worksheet_template_2_search_view"/>
+    </record>
+    <record id="display_stock_lot_list_view_with_maintenances" model="ir.actions.act_window">
+        <field name="name">Stock lot custom view</field>
```
vava-odoo:
```suggestion
        <field name="name">Stock lot for HVAC</field>
```

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo>
+    <record id="update_so_line_quantity" model="ir.actions.server">
+        <field name="name">Update SO line Quantity</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+if record.order_id.state != "cancel" and record.order_id.state != "sale":
```
vava-odoo: Why not move this condition in the filter_domain?
Also, shouldn't you care about batch edition?

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,34 @@
...
+if record.order_id.state != "cancel" and record.order_id.state != "sale": 
+    record['product_uom_qty'] = len(record.x_maintained_serials)
+]]></field>
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+    </record>
+    <record id="update_fs_task_serials_on_delivery" model="ir.actions.server">
+        <field name="name">Update FS Task Serials on Delivery</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for task in record.env['project.task'].search([('sale_order_id.id', '=', record.sale_id.id)]):
```
vava-odoo:
```suggestion
for task in record.env['project.task'].search([('sale_order_id', '=', record.sale_id)]):
```

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,34 @@
...
+    if task.is_fsm:
```
vava-odoo: Add that to the domain above

hvac_services/data/ir_actions_server.xml
vava-odoo: same remark about batch-edit

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,34 @@
...
+    </record>
+    <record id="update_fs_task_serials_on_delivery" model="ir.actions.server">
+        <field name="name">Update FS Task Serials on Delivery</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for task in record.env['project.task'].search([('sale_order_id.id', '=', record.sale_id.id)]):
+    if task.is_fsm: 
+        for move in record.move_ids_without_package:
+            if move.product_id.is_storable and move.product_id.tracking == "serial" and move.lot_ids not in task.x_serials:
+                task['x_serials'] += move.lot_ids
```
vava-odoo: Does that do what you want? Shouldn't you use Command.link?

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,34 @@
...
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+    </record>
+    <record id="update_fs_task_serials_on_delivery" model="ir.actions.server">
+        <field name="name">Update FS Task Serials on Delivery</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for task in record.env['project.task'].search([('sale_order_id.id', '=', record.sale_id.id)]):
+    if task.is_fsm: 
+        for move in record.move_ids_without_package:
+            if move.product_id.is_storable and move.product_id.tracking == "serial" and move.lot_ids not in task.x_serials:
```
vava-odoo: Last condition is weird. Don't you want to check that all lot from lot_ids are in x_serials? Wouldn't be easier to list all that are not in, and add only those?

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,34 @@
...
+                task['x_serials'] += move.lot_ids
+]]></field>
+        <field name="model_id" ref="stock.model_stock_picking"/>
+    </record>
+    <record id="add_sale_order_line_serials_to_sale_order" model="ir.actions.server">
+        <field name="name">Set x_maintained_serials of the sale order equal to the serials of its order lines</field>
```
vava-odoo: Let's keep this non-technical
```suggestion
        <field name="name">Set serials of the sale order equal to the serials of its order lines</field>
```

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,34 @@
...
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for line in record.order_line:
+    if line.x_maintained_serials not in record.x_maintained_serials:
+        record['x_maintained_serials'] += line.x_maintained_serials
```
vava-odoo: I wonder... Wouldn't a related field do the job? like `order_line.x_maintained_serials`

hvac_services/data/ir_model_access.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo>
+    <record id="industry_fsm_report.x_project_task_worksheet_template_2_ir_model_access_2" model="ir.model.access" forcecreate="1">
+        <field name="name">x_project_task_worksheet_template_2_user_access</field>
+        <field name="group_id" ref="project.group_project_user"/>
+        <field name="model_id" ref="industry_fsm_report.x_project_task_worksheet_template_2"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
```
vava-odoo: Is it wanted to allow same access rights to manager and user? If yes, no need to define it for manager, as a manager is a user and get access from this.

hvac_services/data/ir_model_fields.xml
```diff
@@ -0,0 +1,216 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo>
+    <record id="x_maintained_serials_on_sale_order_line_field" model="ir.model.fields">
+        <field name="name">x_maintained_serials</field>
+        <field name="ttype">many2many</field>
+        <field name="field_description">Serials</field>
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="relation">stock.lot</field>
+        <field name="copied" eval="True"/>
+        <field name="on_delete" eval="False"/>
```
vava-odoo: on_delete only makes sense on m2o
```suggestion
```

hvac_services/data/ir_model_fields.xml
```diff
@@ -0,0 +1,216 @@
...
+        <field name="copied" eval="True"/>
+        <field name="model_id" ref="product.model_product_template"/>
+        <field name="on_delete" eval="False"/>
+    </record>
+    <record id="x_product_is_maintenance_field" model="ir.model.fields">
+        <field name="name">x_product_is_maintenance</field>
+        <field name="ttype">boolean</field>
+        <field name="field_description">Is Maintenance or Repair</field>
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="related">product_template_id.x_is_maintenance</field>
```
vava-odoo: Isn't it product_id on a SO line?
```suggestion
        <field name="related">product_id.x_is_maintenance</field>
```

hvac_services/data/ir_model_fields.xml
```diff
@@ -0,0 +1,216 @@
...
+        <field name="model_id" ref="stock.model_stock_lot"/>
+        <field name="readonly" eval="True"/>
+        <field name="store" eval="False"/>
+        <field name="compute"><![CDATA[
+for record in self: 
+    x_delivery = False
+    for delivery in record.delivery_ids:
+        if delivery.picking_type_code == "outgoing" and (x_delivery == False or x_delivery < delivery.date_done): 
+            x_delivery = delivery.date_done
+    record['x_delivery'] = x_delivery
```
vava-odoo: Couldn't it be a one-liner? Something like (untested)
```suggestion
    record['x_delivery'] = max(record.delivery_ids.filtered(lambda d: d.type_code == "outgoing").mapped('date_done'), default=False)
```

hvac_services/data/ir_model_fields.xml
```diff
@@ -0,0 +1,216 @@
...
+        <field name="copied" eval="True"/>
+        <field name="field_description">Maintenances</field>
+        <field name="model_id" ref="stock.model_stock_lot"/>
+        <field name="relation">sale.order</field>
+        <field name="store" eval="False"/>
+        <field name="compute"><![CDATA[
+for record in self:
+    record['x_maintenances'] = self.env['sale.order.line'].search([
+        ('product_template_id', '=', self.env.ref('hvac_services.product_template_maintenance')), 
+        ('x_maintained_serials', '=', record.id)]).order_id
```
vava-odoo: What if there are two lines linked to different orders? Maybe more robust with 
```suggestion
    record['x_maintenances'] = [Command.link(id) for id in (self.env['sale.order.line'].search([
        ('product_template_id', '=', self.env.ref('hvac_services.product_template_maintenance')), 
        ('x_maintained_serials', 'in', record.id),
    ]).mapped('order_id.ids'))
```

hvac_services/data/ir_model_fields.xml
```diff
@@ -0,0 +1,216 @@
...
+        <field name="store" eval="False"/>
+        <field name="compute"><![CDATA[
+for record in self:
+    x_maintenance = 'N/A'
+    if len(record.delivery_ids) > 0 :
+        x_maintenance = 'No Maintenance'
+    for maintenance in record.x_maintenances:
+        if maintenance.subscription_state == '3_progress': 
+            x_maintenance = 'Maintained'
+            break
```
vava-odoo:
```suggestion
    if any(maintenance.subscription_state == '3_progress' for maintenance in record.x_maintenances)
        x_maintenance = 'Maintained'
```

hvac_services/data/ir_models.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo>
+    <record id="industry_fsm_report.x_project_task_worksheet_template_2" model="ir.model" forcecreate="1">
```
vava-odoo: Same remark: why use a model from industry_fsm and not a new one?

hvac_services/data/ir_ui_view.xml
```diff
@@ -0,0 +1,333 @@
...
+    <record id="product_template_form_view" model="ir.ui.view">
+        <field name="name">Custom form view for Product Template</field>
+        <field name="inherit_id" ref="product.product_template_only_form_view"/>
+        <field name="mode">extension</field>
+        <field name="model">product.template</field>
+        <field name="priority">160</field>
+        <field name="type">form</field>
+        <field name="active" eval="True"/>
+        <field name="arch" type="xml">
+            <data>
```
vava-odoo: Useless
```suggestion
```

hvac_services/data/ir_ui_view.xml
```diff
@@ -0,0 +1,333 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo>
+    <record id="product_template_form_view" model="ir.ui.view">
+        <field name="name">Custom form view for Product Template</field>
```
vava-odoo:
```suggestion
        <field name="name">HVAC form view for Product Template</field>
```

hvac_services/data/ir_ui_view.xml
```diff
@@ -0,0 +1,333 @@
...
+        <field name="name">Custom form view for Sale Order</field>
+        <field name="inherit_id" ref="sale.view_order_form"/>
+        <field name="mode">extension</field>
+        <field name="model">sale.order</field>
+        <field name="priority">160</field>
+        <field name="type">form</field>
+        <field name="active" eval="True"/>
+        <field name="arch" type="xml">
+            <data>
+                <xpath expr="//form[1]/sheet[1]/notebook[1]/page[@name='order_lines']/field[@name='order_line']/list[1]/field[@name='name']" position="after">
```
vava-odoo: Can't you reduce the xpath to make it more upgrade-friendly, something like
```suggestion
                <xpath expr="//page[@name='order_lines']/field[@name='order_line']//field[@name='name']" position="after">
```

hvac_services/data/ir_ui_view.xml
```diff
@@ -0,0 +1,333 @@
...
+                <xpath expr="//button[@name='action_view_stock_valuation_layers']" position="after">
+                    <button class="oe_stat_button" icon="fa-rotate-right" type="action" name="%(get_maintenances_action)d">
+                        <field widget="statinfo" name="x_sale_order_count" string="Maintenances"/>
+                    </button>
+                    <button class="oe_stat_button" icon="fa-flash" type="action" name="%(get_field_services_action)d">
+                        <field widget="statinfo" name="x_project_task_count" string="Field Services"/>
+                    </button>
+                </xpath>
+                <xpath expr="//field[@name='partner_ids']" position="after">
+                    <field name="x_delivery" force_save="1" readonly="True" widget="remaining_days" decoration-danger="x_delivery &lt; current_date"/>
```
vava-odoo: Does the decoration work on a form?

hvac_services/data/ir_ui_view.xml
```diff
@@ -0,0 +1,333 @@
...
+        <field name="name">Custom form view for project task</field>
+        <field name="inherit_id" ref="project.view_task_form2"/>
+        <field name="mode">extension</field>
+        <field name="model">project.task</field>
+        <field name="priority">9990</field>
+        <field name="type">form</field>
+        <field name="active" eval="True"/>
+        <field name="arch" type="xml">
+            <data>
+                <xpath expr="//form[1]/sheet[1]/notebook[1]" position="inside">
```
vava-odoo:
```suggestion
                <xpath expr="//notebook[1]" position="inside">
```

hvac_services/data/ir_ui_view.xml
```diff
@@ -0,0 +1,333 @@
...
+        <field name="inherit_id" ref="project.view_task_tree2"/>
+        <field name="mode">extension</field>
+        <field name="model">project.task</field>
+        <field name="priority">160</field>
+        <field name="type">list</field>
+        <field name="active" eval="True"/>
+        <field name="arch" type="xml">
+            <data>
+                <xpath expr="//list[1]" position="attributes">
+                <attribute name="default_group_by"/>
```
vava-odoo, line 171: indent issue

hvac_services/data/ir_ui_view.xml
```diff
@@ -0,0 +1,333 @@
...
+            <search>
+                <field name="x_name"/>
+                <filter string="Created on" date="create_date" name="create_date"/>
+                <filter name="group_by_month" string="Created on" context="{'group_by': 'create_date:month'}"/>
+            </search>
+        </field>
+    </record>
+    <record id="x_project_task_worksheet_template_2_form_view_extended" model="ir.ui.view">
+        <field name="name">Odoo Studio: template_view_INST/MAIN/REPA_Worksheet customization</field>
+        <field name="inherit_id" ref="industry_fsm_report.x_project_task_worksheet_template_2_form_view"/>
```
vava-odoo: You inherit from a view that is defined just above... That is due to the historical edition steps of the view with Studio, but we don't want that here. Move the useful part in the original view definition above

hvac_services/data/ir_ui_view_2.xml
vava-odoo, line 1: Is there any particular reason to have it in a separate file?

hvac_services/data/knowledge_article.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <template id="welcome_article_body">
+<h1 class="oe-hint" data-oe-version="2.0">HVAC Services</h1><p>Hello there!</p><p>You just installed the Odoo for HVAC Services package. By doing so, we have installed a bunch of necessary apps to run your bar efficiently.</p><p>Discover the basics of this package and explore all the possibilities Odoo offers to improve your experience.</p><div class="o_editor_banner user-select-none lh-1 d-flex align-items-center alert alert-info pb-0 pt-3 o-contenteditable-false" data-oe-role="status">
```
vava-odoo: Could you make it readable?

hvac_services/data/mail_message.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record model="mail.message" id="notification_knowledge">
+        <field name="model">discuss.channel</field>
+        <field name="res_id" ref="mail.channel_all_employees"/>
+        <field name="message_type">email</field>
+        <field name="author_id" ref="base.partner_root"/>
+        <field name="subtype_id" ref="mail.mt_comment"/>
+        <field name="subject">🚀 Get started with Odoo Hvac Services Shop</field>
```
vava-odoo:
```suggestion
        <field name="subject">🚀 Get started with Odoo Hvac Services</field>
```

hvac_services/data/knowledge_tour.xml
vava-odoo, line 1: remove

hvac_services/data/product_product.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="product_product_installation" model="product.product">
+        <field name="product_tmpl_id" ref="product_template_installation"/>
+        <field name="is_favorite" eval="True"/>
```
vava-odoo: Don't you define is_favorite in the template already?

hvac_services/data/res_config_settings.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="res_config_settings_enable" model="res.config.settings">
+        <field name="group_sale_order_template" eval="True"/>
+        <field name="group_stock_production_lot" eval="True"/>
+        <field name="group_project_recurring_tasks" eval="True"/>
+        <field name="module_industry_fsm_report" eval="True"/>
```
vava-odoo: Is it necessary? I guess if industry_fsm_sale_report is installed, this module is as well
```suggestion
```

hvac_services/demo/hr_employee.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="hr.employee_admin" model="hr.employee" forcecreate="1">
```
vava-odoo, line 3: If it happens to be created again with forcecreate, I bet you'd need some required fields defined here, like the name, or it would fail

hvac_services/demo/hr_job.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="hr_job_technician" model="hr.job">
```
vava-odoo: Could be in data/

hvac_services/demo/product_pricelist.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="product_pricelist_1" model="product.pricelist">
```
vava-odoo: This one is generated automatically, you can remove (and references to it as well)

hvac_services/demo/product_template.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="product_template_4" model="product.template" context="{'create_product_product': False}">
```
vava-odoo: All could be in data/

hvac_services/demo/purchase_order_post.xml
```diff
@@ -0,0 +1,118 @@
...
+    <function model="purchase.order" name="button_confirm">
+        <value eval="[
+            ref('purchase_order_1'),
+            ref('purchase_order_2')]"/>
+    </function>
+    <function name="write" model="stock.move.line">
+        <value model="stock.move.line" eval="obj().search([
+        ('product_id', '=', ref('product_product_6')),
+        ('picking_id', '=', obj().env.ref('hvac_services.purchase_order_2').picking_ids.id)])[0].id"/>
+        <value model="stock.lot" eval="{'lot_id': ref('stock_lot_1'), 'lot_name': obj().env.ref('hvac_services.stock_lot_1').name}"/>
```
vava-odoo: lot_name is not necessary if I recall correctly
```suggestion
        <value model="stock.lot" eval="{'lot_id': ref('stock_lot_1')}"/>
```

hvac_services/demo/purchase_order_post.xml
```diff
@@ -0,0 +1,118 @@
...
+<odoo noupdate="1">
+    <function model="purchase.order" name="button_confirm">
+        <value eval="[
+            ref('purchase_order_1'),
+            ref('purchase_order_2')]"/>
+    </function>
+    <function name="write" model="stock.move.line">
+        <value model="stock.move.line" eval="obj().search([
+        ('product_id', '=', ref('product_product_6')),
+        ('picking_id', '=', obj().env.ref('hvac_services.purchase_order_2').picking_ids.id)])[0].id"/>
```
vava-odoo: To avoid the slice, you could limit the search to 1, and add a criterion that lot_id is False in the domain (same as in it_hardware for reference)

hvac_services/demo/res_partner.xml
```diff
@@ -0,0 +1,205 @@
...
+    <record id="res_partner_33" model="res.partner">
+        <field name="name">Third Company</field>
+        <field name="country_id" ref="base.be"/>
+        <field name="street">Rue de Ramilies 1</field>
+        <field name="zip">1367</field>
+        <field name="city">Ramilies</field>
+        <field name="is_company" eval="True"/>
+        <field name="image_1920" type="base64" file="hvac_services/static/src/binary/res_partner/22-image_1920"/>
+    </record>
+    <record id="res_partner_34" model="res.partner">
```
vava-odoo: Remove, use the ones defined in base_industry_data, and use those instead (only edit if need be)

hvac_services/demo/res_users.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="res_users_1" model="res.users">
+        <field name="partner_id" ref="res_partner_1"/>
+        <field name="login">andrew.blake@hvac_services.example.com</field>
+        <field name="company_ids" eval="[(6, 0, [ref('base.main_company')])]"/>
+        <field name="company_id" ref="base.main_company"/>
```
vava-odoo:
```suggestion
```

hvac_services/demo/sale_order_2.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="sale_order_5" model="sale.order">
+        <field name="x_maintained_serials" eval="[(6, 0, [ref('stock_lot_14')])]"/>
```
vava-odoo: Isn't it computed by the server action?

hvac_services/demo/sale_order_line.xml
```diff
@@ -0,0 +1,238 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo auto_sequence="1" noupdate="1">
+    <record id="sale_order_line_38" model="sale.order.line">
+        <field name="order_id" ref="sale_order_11"/>
+        <field name="product_id" ref="product_product_repair"/>
+        <field name="analytic_distribution" eval="{'2': 100}"/>
```
vava-odoo: 2 is related to an id from the db. I'm not sure you need to define the distribution, but if you do, you'll need use a ref instead of an hardcoded id

hvac_services/demo/sale_order_template.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="sale_order_template_2" model="sale.order.template">
+        <field name="name">Installation (S12ET)</field>
```
vava-odoo: Why this template is in demo? I bet you could move it in data as well

hvac_services/demo/x_project_task_worksheet_template_2.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="x_project_task_worksheet_template_2_1" model="x_project_task_worksheet_template_2">
+        <field name="x_name">S00001 - Installation</field>
+        <field name="x_comments"><![CDATA[<div data-oe-version="2.0">All went fine</div>]]></field>
+        <field name="x_worker_signature" type="base64" file="hvac_services/static/src/binary/x_project_task_worksheet_template_2/1-x_worker_signature"/>
+        <field name="x_refrigerant_product" ref="product_template_9"/>
+        <field name="x_product_is_functional" eval="True"/>
+        <field name="x_operation">Installation</field>
+        <field name="x_date">2025-06-08</field>
```
vava-odoo: No hardcoded date. Either the default one is fine and you remove, either you add a relative date

hvac_services/static/description/index.html
```diff
@@ -0,0 +1,29 @@
+<h2><strong>Odoo for HVAC services</strong></h2>
+<p>This industry package includes all the essential features to manage your handyman business, from customer relationship management to field service operations and invoicing.</p>
```
vava-odoo: handyman? Probably forgot to remove this line :smile: 
```suggestion
```

hvac_services/static/src/js/my_tour.js
vava-odoo, line 1: use the one from base_industry_data

vava-odoo (commented):
Hey @frva-odoo 
Thanks for this! 
I wrote quite a few comments, but don't be afraid, nothing really blocking this from moving on.
Don't hesitate to come back to me if any of these is unclear.
Good luck :muscle:

hvac_services/data/ir_actions_act_window.xml
frva-odoo: What would be the correct way to write this domain ? I saw "_ids '=' _id" a few times and it works as a 'contains', but I couldn't find an easy way to write it other than that.

hvac_services/data/ir_actions_act_window_2.xml
frva-odoo: Not an error, it is here so that the user can open the form view from this list. But the id and name does not reflect that so I'll change them

hvac_services/data/ir_actions_server.xml
frva-odoo: You have to check all the lines of the sale order and add their serials if they have one, so I don't think it's possible to do that with a related field

hvac_services/data/ir_ui_view_2.xml
frva-odoo: Yes, the view references an action that uses another view referencing another action. That's why there is 2 files for ir_ui_view and 2 files for ir_act_windows

hvac_services/__manifest__.py
```diff
@@ -0,0 +1,67 @@
...
+        'demo/purchase_order_post.xml',
+        'demo/project_task.xml',
+        'demo/sale_order_post.xml',
+        'demo/x_project_task_worksheet_template.xml',
+    ],
+    'license': 'OPL-1',
+    'author': 'Odoo S.A.',
+    "cloc_exclude": [
+        "data/knowledge_article.xml",
+        "data/ir_ui_view.xml",
```
vava-odoo: please split between qweb and other views to exclude only qweb views.

hvac_services/data/base_automation.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo>
+    <record id="base_automation_SO_line" model="base.automation">
+        <field name="name">On SO line Serial Edit</field>
+        <field name="action_server_ids" eval="[(6, 0, [ref('update_so_line_quantity')])]"/>
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="filter_domain">[("product_id.x_is_maintenance", "!=", False), ("order_id.state", "in", ["draft", "sent"])]</field>
```
vava-odoo, line 8: If I translate correctly the condition that was in the server action, you lost the negation, right?
```suggestion
        <field name="filter_domain">[("product_id.x_is_maintenance", "!=", False), ("order_id.state", "not in", ["draft", "sent"])]</field>
```
It probably requires tests :see_no_evil:

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,46 @@
...
+        task['x_serials'] = [Command.link(lot_id.id)]
+]]></field>
+        <field name="model_id" ref="project.model_project_task"/>
+    </record>
+    <record id="update_fs_task_serials_on_delivery" model="ir.actions.server">
+        <field name="name">Update FS Task Serials on Delivery</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for stock_pick in records:
+    for task in stock_pick.env['project.task'].search([('sale_order_id', '=', stock_pick.sale_id.id), ('is_fsm', '=', True)]):
```
vava-odoo: Didn't you forget to remove the .id?
```suggestion
    for task in stock_pick.env['project.task'].search([('sale_order_id', '=', stock_pick.sale_id), ('is_fsm', '=', True)]):
```

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,46 @@
...
+        <field name="model_id" ref="stock.model_stock_picking"/>
+    </record>
+    <record id="add_sale_order_line_serials_to_sale_order" model="ir.actions.server">
+        <field name="name">Set serials of the sale order equal to the serials of its order lines</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for so in records:
+    for line in so.order_line:
+        if line.x_maintained_serials_ids not in so.x_maintained_serials_ids:
+            so['x_maintained_serials_ids'] = [Command.link(line.x_maintained_serials_ids)]
```
vava-odoo: Does that work with ids? I think you need to loop

hvac_services/data/ir_model_fields.xml
```diff
@@ -0,0 +1,206 @@
...
+        <field name="name">x_maintenances</field>
+        <field name="ttype">many2many</field>
+        <field name="copied" eval="True"/>
+        <field name="field_description">Maintenances</field>
+        <field name="model_id" ref="stock.model_stock_lot"/>
+        <field name="relation">sale.order</field>
+        <field name="store" eval="False"/>
+        <field name="compute"><![CDATA[
+for record in self:
+    record['x_maintenances'] = [order_id.id for order_id in (self.env['sale.order.line'].search([
```
vava-odoo: Didn't you miss the Command.link?
```suggestion
    record['x_maintenances'] = [Command.link(order_id.id) for order_id in self.env['sale.order.line'].search([
```

hvac_services/data/ir_ui_view.xml
```diff
@@ -0,0 +1,302 @@
...
+                        <field name="x_operation" widget="selection_badge" required="True"/>
+                        <field name="x_date" string="Date" required="True"/>
+                    </group>
+                    <group string="Repair" invisible="x_operation != &quot;Repair&quot;">
+                        <field name="x_primary_root_cause"/>
+                        <field name="x_root_cause_comments" help="Provide details if additional root cause led to the repair operation"/>
+                    </group>
+                    <group string="Refrigerant">
+                        <field name="x_refrigerant_product" 
+                            domain="[('x_is_refrigerant', '!=', False), (&quot;categ_id&quot;, &quot;=&quot;, 5)]"
```
vava-odoo: 5 is the id for this clean db, but will probably be different in any other user db. Please use a ref.
And replace all `&quot;` by `'` please

vava-odoo: reading twice the diff, the domain was replaced in the attribute, so I don't think you need to keep the second part
```suggestion
                            domain="[('x_is_refrigerant', '!=', False)" 
```

hvac_services/data/ir_actions_act_window_2.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo>
+    <record id="x_project_task_worksheet_template_ir_actions_get_worksheets" model="ir.actions.act_window">
+        <field name="name">Worksheets</field>
+        <field name="context">{'edit': False, 'create': False, 'delete': False, 'duplicate': False, 'action_xml_id': 'industry_fsm_report.fsm_worksheets_action_settings', 'worksheet_template_id': 2}</field>
```
vava-odoo: Shouldn't it be something like
```suggestion
        <field name="context">{'edit': False, 'create': False, 'delete': False, 'duplicate': False, 'action_xml_id': 'industry_fsm_report.fsm_worksheets_action_settings', 'worksheet_template_id': ref('worksheet_template')}</field>
```

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,46 @@
...
+]]></field>
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+    </record>
+    <record id="update_fs_task_serials_on_create" model="ir.actions.server">
+        <field name="name">Update FS Task Serials on create</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for task in records:
+    for lot_id in task.sale_order_id.order_line.mapped('x_maintained_serials_ids').filtered(lambda l: l and l not in task.x_serials):
+        task['x_serials'] = [Command.link(lot_id.id)]
```
vava-odoo, line 18: Thinking about it: what if we remove an order line? You probably need to clear it first, no?
```suggestion
    task['x_serials'] = [Command.clear()]
    for lot_id in task.sale_order_id.order_line.mapped('x_maintained_serials_ids').filtered(lambda l: l and l not in task.x_serials):
        task['x_serials'] = [Command.link(lot_id.id)]
```

hvac_services/data/ir_actions_server.xml
vava-odoo: Also I'm a bit concerned about the performances of such method. Having a search in a loop is not at all recommanded.

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,46 @@
...
+    <record id="update_fs_task_serials_on_delivery" model="ir.actions.server">
+        <field name="name">Update FS Task Serials on Delivery</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for stock_pick in records:
+    for task in stock_pick.env['project.task'].search([('sale_order_id', '=', stock_pick.sale_id.id), ('is_fsm', '=', True)]):
+        for move in stock_pick.move_ids_without_package:
+            if move.product_id.is_storable and move.product_id.tracking == "serial":
+                for lot_id in move.lot_ids.filtered(lambda l: l not in task.x_serials):
+                    task['x_serials'] = [Command.link(lot_id.id)]
```
vava-odoo, line 31: And same remark as above, don't you need to clear it first?

hvac_services/data/ir_model_fields.xml
```diff
@@ -0,0 +1,206 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo>
+    <record id="x_maintained_serials_on_sale_order_line_field" model="ir.model.fields">
+        <field name="name">x_maintained_serials_ids</field>
+        <field name="ttype">many2many</field>
+        <field name="field_description">Serials</field>
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="relation">stock.lot</field>
+        <field name="copied" eval="True"/>
```
vava-odoo, line 9: I wonder about those copied values... If we duplicate a sale order line, do we want to keep the serials? (real question)
Could you check other ones in the file and see if relevant or not?

hvac_services/data/ir_ui_view.xml
```diff
@@ -0,0 +1,302 @@
...
+        <field name="priority">160</field>
+        <field name="type">list</field>
+        <field name="active" eval="True"/>
+        <field name="arch" type="xml">
+            <xpath expr="//list[1]" position="attributes">
+                <attribute name="default_group_by"/>
+                <attribute name="default_order">planned_date_begin desc</attribute>
+            </xpath>
+            <xpath expr="//field[@name='user_ids']" position="after">
+            <xpath expr="//field[@name='date_deadline']" position="move"/></xpath>
```
vava-odoo:
```suggestion
            <xpath expr="//field[@name='user_ids']" position="after">
                <xpath expr="//field[@name='date_deadline']" position="move"/>
            </xpath>
```

hvac_services/data/ir_ui_view_2.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="view_partner_stock_form" model="ir.ui.view">
```
vava-odoo: I still doubt you need a second file for this view only

hvac_services/data/knowledge_article.xml
```diff
@@ -0,0 +1,421 @@
...
+                </p>
+            </div>
+        </div>
+        <p>
+            <br />
+        </p>
+        <div class="oe-hint"
+            style='box-sizing: border-box; color: rgb(55, 65, 81); font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Ubuntu, "Noto Sans", Arial, "Odoo Unicode Support Noto", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;'>
+            <br />
+        </div>
```
vava-odoo: Looks empty
```suggestion
```

hvac_services/data/product_product.xml
```diff
@@ -0,0 +1,48 @@
...
+    </record>
+    <record id="product_product_maintenance" model="product.product">
+        <field name="product_tmpl_id" ref="product_template_maintenance"/>
+    </record>
+    <record id="product_product_repair" model="product.product">
+        <field name="product_tmpl_id" ref="product_template_repair"/>
+    </record>
+    <record id="appointment_account_payment.default_booking_product" model="product.product" forcecreate="1">
+        <field name="product_tmpl_id" ref="appointment_account_payment.default_booking_product_product_template"/>
+    </record>
```
vava-odoo: This product is protected against deletion
```suggestion
```

hvac_services/data/product_product.xml
```diff
@@ -0,0 +1,48 @@
...
+    <record id="product_product_4" model="product.product">
+        <field name="product_tmpl_id" ref="product_template_4"/>
+        <field name="standard_price">100.0</field>
+    </record>
+    <record id="industry_fsm_sale.field_service_product" model="product.product" forcecreate="1">
+        <field name="product_tmpl_id" ref="industry_fsm_sale.field_service_product_product_template"/>
+    </record>
```
vava-odoo: This is useless. Either the product template exists, and the product product as well. Either it doesn't and this will fail because you miss required fields to create a new product. I'd just drop that part
```suggestion
```

hvac_services/demo/product_supplierinfo.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="product_supplierinfo_5" model="product.supplierinfo">
+        <field name="partner_id" ref="base_industry_data.res_partner_33"/>
+        <field name="product_tmpl_id" ref="product_template_8"/>
```
vava-odoo: I guess product_id is sufficient here,
```suggestion
```

vava-odoo (changes requested):
Good improvement, still a few comments to address before moving on.
Good luck!

hvac_services/data/base_automation.xml
frva-odoo: I also used the other states, instead of saying "state != 'cancel' and state != 'sale'", I said "state in ['draft','sent']". A bit confusing to change like that but it works the same way

hvac_services/data/ir_model_fields.xml
frva-odoo: I was not sure either so I asked the PO and after discussing it, it is pertinent. I still looked at the other ones to be sure.

hvac_services/data/ir_actions_server.xml
frva-odoo: It shouldn't happen as the task is created when the sale order is confirmed, and you can't remove a line from it when confirmed. Additionally, the x_serials fields has all serials from the sale order lines and from the deliveries done linked to the so and we also want to be able to add others manually, so clearing it like that would remove serials that we want.

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,47 @@
...
+        task['x_serials'] = [Command.link(lot_id.id)]
+]]></field>
+        <field name="model_id" ref="project.model_project_task"/>
+    </record>
+    <record id="update_fs_task_serials_on_delivery" model="ir.actions.server">
+        <field name="name">Update FS Task Serials on Delivery</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for stock_pick in records.filtered(lambda s: s.sale_id):
+    for task in stock_pick.sale_id.tasks_ids:
```
vava-odoo: filtered on sale_id seems useless: if none, the loop ends directly;
but you lost the filtered on fsm tasks only, no?
```suggestion
for stock_pick in records:
    for task in stock_pick.sale_id.tasks_ids.filtered('is_fsm'):
```

hvac_services/data/ir_actions_server.xml
vava-odoo: OK thanks for the explanation. 
maybe Command.set would be the best, no? I feel it's clearer.
Also, do you need the map on m2m.m2m ? I think the ORM handles that 
Could you try this? Also in the case of no ids to see if it doesn't crash.
```suggestion
    task['x_serials'] = [Command.set(task.sale_order_id.order_line.x_maintained_serials_ids.filtered(lambda l: l and l not in task.x_serials).ids]
```

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,47 @@
...
+    </record>
+    <record id="add_sale_order_line_serials_to_sale_order" model="ir.actions.server">
+        <field name="name">Set serials of the sale order equal to the serials of its order lines</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for so in records:
+    for line in so.order_line:
+        if line.x_maintained_serials_ids not in so.x_maintained_serials_ids:
+            for serial in line.x_maintained_serials_ids:
+                so['x_maintained_serials_ids'] = [Command.link(serial.id)]
```
vava-odoo: Isn't it simpler that way? I feel like it's clearer and equivalent, but tell me if I miss something
```suggestion
    so['x_maintained_serials_ids'] = [Command.set(so.order_line.x_maintained_serials_ids.ids)]
```
And to me, this is ... a related field :slightly_smiling_face: EDIT ah no, related would only take the first :pensive:

hvac_services/data/ir_actions_server.xml
```diff
@@ -0,0 +1,47 @@
...
+    <record id="update_fs_task_serials_on_delivery" model="ir.actions.server">
+        <field name="name">Update FS Task Serials on Delivery</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for stock_pick in records.filtered(lambda s: s.sale_id):
+    for task in stock_pick.sale_id.tasks_ids:
+        for move in stock_pick.move_ids_without_package:
+            if move.product_id.is_storable and move.product_id.tracking == "serial":
+                for lot_id in move.lot_ids.filtered(lambda l: l not in task.x_serials):
+                    task['x_serials'] = [Command.link(lot_id.id)]
```
vava-odoo: Idea
```suggestion
    for task in stock_pick.sale_id.tasks_ids:
        task['x_serials'] = [Command.set(stock_pick.move_ids_without_package.filtered(lambda m: m.product_id.is_storable and m.product_id.tracking == "serial").lot_ids.ids)] if task.is_fsm else False
```

hvac_services/data/base_automation.xml
vava-odoo: Still not convinced here

hvac_services/data/ir_actions_server.xml
vava-odoo: not possible because we want to edit manually

hvac_services/data/ir_actions_server.xml
vava-odoo: same, no set/clear because we allow to modify it manually

vava-odoo (approved):
@robodoo r+

PR #697, opened 2025-06-17
[IMP] it_hardware: improve repair service flow

abdrahmanrashed:
This commit improves the user experience by doing the following:
-  enabling use_product_returns for the customer care helpdesk project.
- enabling use_product_repairs for the customer care helpdesk project.
- Updating "Flow 5: Repair Service" using the newly enabled features to make the flow more functional by making process of returning/repairing products for customers easier for users.

task-4656025

### DIFF
--- a/it_hardware/data/helpdesk_config.xml
+++ b/it_hardware/data/helpdesk_config.xml
@@ -1,0 +1,8 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="helpdesk.helpdesk_team1" model="helpdesk.team" forcecreate="True">
+        <field name="name">Customer Care</field>
+        <field name="use_product_returns" eval="True"/>
+        <field name="use_product_repairs" eval="True"/>
+    </record>
+</odoo>
--- a/it_hardware/demo/purchase_order_confirm.xml
+++ b/it_hardware/demo/purchase_order_confirm.xml
@@ -5,32 +5,37 @@
     <function name="write" model="stock.move.line">
         <value model="stock.move.line" eval="obj().search([
         ('product_id', '=', ref('product_product_5')),
-        ('picking_id', '=', obj().env.ref('it_hardware.purchase_order_1').picking_ids.id)])[0].id"/>
-        <value model="stock.lot" eval="{'lot_id': ref('stock_lot_2'), 'lot_name': obj().env.ref('it_hardware.stock_lot_2').name}"/>
+        ('picking_id', '=', obj().env.ref('it_hardware.purchase_order_1').picking_ids.id),
+        ('lot_id', '=', False)], limit=1).id"/>
+        <value model="stock.lot" eval="{'lot_id': ref('stock_lot_2')}"/>
     </function>
     <function name="write" model="stock.move.line">
         <value model="stock.move.line" eval="obj().search([
         ('product_id', '=', ref('product_product_5')),
-        ('picking_id', '=', obj().env.ref('it_hardware.purchase_order_1').picking_ids.id)])[1].id"/>
-        <value model="stock.lot" eval="{'lot_id': ref('stock_lot_3'), 'lot_name': obj().env.ref('it_hardware.stock_lot_3').name}"/>
+        ('picking_id', '=', obj().env.ref('it_hardware.purchase_order_1').picking_ids.id),
+        ('lot_id', '=', False)], limit=1).id"/>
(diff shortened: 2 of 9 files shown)


it_hardware/data/helpdesk_config.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="helpdesk.helpdesk_team1" model="helpdesk.team" forcecreate="True">
```
vava-odoo, line 3: Please add also the name, otherwise forcecreate will try to recreate the record but will fail since name is required without a default value

it_hardware/demo/purchase_order_confirm.xml
```diff
@@ -74,4 +74,40 @@
         <value model="stock.picking" eval="obj().env.ref('it_hardware.purchase_order_3').picking_ids.ids"/>
     </function>
     <function model="purchase.order" name="action_create_invoice" eval="[ref('purchase_order_3')]"/>
+
+    <function model="purchase.order" name="button_confirm" eval="[ref('purchase_order_4')]"/>
+    <function name="write" model="stock.move.line">
+        <value model="stock.move.line" eval="obj().search([
+        ('product_id', '=', ref('product_product_9')),
+        ('picking_id', '=', obj().env.ref('it_hardware.purchase_order_4').picking_ids.id)])[0].id"/>
+        <value model="stock.lot" eval="{'lot_id': ref('stock_lot_11'), 'lot_name': obj().env.ref('it_hardware.stock_lot_11').name}"/>
```
vava-odoo: (nitpicking) is it mandatory to provide the lot_name? Isn't it automatically retrieved by the lot_id.name?

it_hardware/demo/purchase_order_confirm.xml
```diff
@@ -74,4 +74,40 @@
         <value model="stock.picking" eval="obj().env.ref('it_hardware.purchase_order_3').picking_ids.ids"/>
     </function>
     <function model="purchase.order" name="action_create_invoice" eval="[ref('purchase_order_3')]"/>
+
+    <function model="purchase.order" name="button_confirm" eval="[ref('purchase_order_4')]"/>
+    <function name="write" model="stock.move.line">
+        <value model="stock.move.line" eval="obj().search([
+        ('product_id', '=', ref('product_product_9')),
+        ('picking_id', '=', obj().env.ref('it_hardware.purchase_order_4').picking_ids.id)])[0].id"/>
```
vava-odoo: (nitpicking) Not a big fan of [0], [1], ... Couldn't you add a condition in the domain that there is no linked lot and add a limit=1 in the search? What do you think?

it_hardware/demo/sale_order_confirm.xml
```diff
@@ -2,7 +2,21 @@
...
-    <function model="sale.order" name="action_confirm" eval="[[ref('sale_order_2')]]"/>
+    <function model="sale.order" name="action_confirm" eval="[[
+        ref('sale_order_1'),
+        ref('sale_order_2'),
+    ]]"/>
+
+    <function name="write" model="stock.move.line">
+        <value model="stock.move.line" eval="obj().search([
+        ('product_id', '=', ref('product_product_9')),
+        ('picking_id', '=', obj().env.ref('it_hardware.sale_order_1').picking_ids.id)])[0].id"/>
```
vava-odoo: same remark

it_hardware/data/helpdesk_config.xml
abdrahmanrashed: What name are you exactly referring to?
I cannot seem to replicate the fail you are talking about, Can you please elaborate?

it_hardware/demo/purchase_order_confirm.xml
abdrahmanrashed: I don't know if it is mandatory, but the "action_create_invoice" fails when I remove it.
I also looked up in the code base and it is written like this everywhere so not sure if there is an alternative.

it_hardware/data/helpdesk_config.xml
vava-odoo: - install helpdesk
- remove helpdesk.helpdesk_team1 record
- try to install it_hardware
-> error

To recreate the record, it needs a name. So just take the one from the original record and add it here.

vava-odoo:
@robodoo squash

vava-odoo:
@robodoo r+

PR #750, opened 2025-06-27
[FIX] *: update pot and tx to saas-18.4

vava-odoo:


vava-odoo:
@robodoo override=ci/industry
@robodoo r+

PR #779, opened 2025-07-01
[IMP] tests: hair salon server action test

frva-odoo:
This PR adds tests to verify the good execution of the server action in the hair_salon module.

TASK-4715882

### DIFF
--- a/tests/test_hair_salon/tests/test_action_server.py
+++ b/tests/test_hair_salon/tests/test_action_server.py
@@ -1,0 +1,72 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+import datetime
+from dateutil.relativedelta import relativedelta
+
+from odoo.tests import tagged
+from odoo.tests.common import TransactionCase
+
+
+@tagged('post_install', '-at_install')
+class ActionServerTestCase(TransactionCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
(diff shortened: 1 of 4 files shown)


tests/test_hair_salon/tests/test_action_server.py
```diff
@@ -0,0 +1,37 @@
...
+
+@tagged('post_install', '-at_install')
+class ActionServerTestCase(TransactionCase):
+
+    def test_break_subpart_calendar_event_server_action(self):
+        event = self.env['calendar.event'].create({
+            'name': 'Event test',
+            'partner_ids': [(6, 0, [self.env['res.partner'].create({'name': 'Partner A'}).id])],
+            'start': '2025-07-01 10:00:00',
+            'stop': '2025-07-01 18:00:00',
```
vava-odoo: need relative date

tests/test_hair_salon/tests/test_action_server.py
```diff
@@ -0,0 +1,37 @@
...
+            'x_break_start': 3.0,
```
vava-odoo: Would be nice to use the complete flow: create an appointment.type with a break defined, and then create a calendar.event for it.
And another without a break defined and check that no other calendar.event is created, maybe?
Also, this makes me realize this feature is not showcased at all in the demo data. Will create a task about it.

vava-odoo (commented):
Thanks @frva-odoo

vava-odoo (approved):
@robodoo r+

PR #780, opened 2025-07-01
[FIX] beverage_distributor: use images from attachment records

dhrs-odoo:
In this PR we have added the images references from the attachment records.

Task-4910794


vava-odoo (approved):
@robodoo r+

PR #784, opened 2025-07-03
[FIX] industry_restaurant: fix error tax not found

frva-odoo:
The module can't be loaded due to the taxes on the demo pos order line being not found. This commit removes those lines as they should not be set directly.


vava-odoo (approved):
@robodoo r+

PR #802, opened 2025-07-08
[FIX] *: Theme installation should uninstall previous one

arthurrrm:
Before this commit, installing a new theme with button_immediate_install leads to errors due to the fact that it doesn't uninstall previous one (which is the default behaviour from the UI). This conflicts with what was defined in the website record.
This commit makes sure the method button_choose_theme is used instead of button_immediate_install to import themes. And the manifest of affected industries has been updated to correct the import of website.xml files which should happen after the theme choice.

task-4920348
opw-4890466


agriculture_shop/demo/website_theme_apply.xml
```diff
@@ -1,11 +1,6 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo>
-    <function name="button_immediate_install" model="ir.module.module" eval="[ref('base.module_theme_enark', raise_if_not_found=False)]"/>
-
-    <function name="_theme_load" model="ir.module.module" context="{'apply_new_theme': True}">
-        <value model="ir.module.module" eval="obj().env['ir.module.module'].search([('name', '=', 'theme_enark')]).ids"/>
-        <value model="ir.module.module" eval="obj().env.ref('website.default_website')"/>
-    </function>
+    <function name="button_choose_theme" model="ir.module.module" eval="[ref('base.module_theme_enark', raise_if_not_found=False)]"/>
```
vava-odoo: Problem is, if design-themes are not in your addons-path, you get an error:
ValueError: Expected singleton: ir.module.module()

vava-odoo:
@robodoo squash

vava-odoo:
@robodoo r+

PR #834, opened 2025-07-15
[FIX] student_organization: fix noupdate attribute

amas-odoo:
fix noupdate attributes on a few records caught by the runbot


vava-odoo:
@robodoo r+ fw=no squash

PR #852, opened 2025-07-17
[FIX] sports_club: fix appointment link on the website

dhrs-odoo:
Before this commit, clicking the Book a Court button on the website directs to a Page not found error because the appointments are not public but accessible only through an invite code, as used by other menus. This commit adapts the link behind the button.

Task-4950786


vava-odoo:
@robodoo r+ squash

PR #869, opened 2025-07-18
[FIX] *: fix broken navigation on website

dhrs-odoo:
In this PR we have fixed the broken navigation on website for several industries.

Task-4953568

### DIFF
--- a/surveyor/demo/website_menu.xml
+++ b/surveyor/demo/website_menu.xml
@@ -6,7 +6,7 @@
         <field name="url">/our-services</field>
         <field name="parent_id" model="website.menu" eval="obj().search([
             ('website_id', '=', ref('website.default_website')),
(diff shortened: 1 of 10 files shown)


surveyor/demo/website_menu.xml
```diff
@@ -6,7 +6,7 @@
         <field name="url">/our-services</field>
         <field name="parent_id" model="website.menu" eval="obj().search([
             ('website_id', '=', ref('website.default_website')),
-            ('url', '=', '/default-main-menu'),
+            ('url', '=', '#'),
```
vava-odoo, line 9: surveyor seems to be the only one in the list where the menu was present before your change. Could you investigate why it is not needed?

vava-odoo (approved):
Thanks @dhrs-odoo

PR #880, opened 2025-07-18
[FIX] hvac_services: fix issue with stages and email

vava-odoo:
This commit fixes 3 small issues of HVAC module:
- handle case where project stages from FSM do not exist;
- realistic email addresses;
- correct template for worksheet.


vava-odoo (commented):
@robodoo r+

PR #885, opened 2025-07-22
[FIX] hvac_services: fix worksheet action context error

frva-odoo:
This PR fixes the context in the x_project_task_worksheet_template_ir_actions_get_worksheets action that had bad syntax.


vava-odoo (approved):
Thanks!
@robodoo r+

PR #905, opened 2025-07-25
[I18N] *: add config file for weblate

dylankiss:


dylankiss:
@robodoo fw=no

vava-odoo (approved):
@robodoo delegate+
@robodoo up to master
Can't we have it in master already, like tx/config? I guess it's harmless (but useful)

dylankiss:
@robodoo r+

PR #912, opened 2025-07-28
[FIX] industry_restaurant: small fix and demo data

dhrs-odoo:
In this PR we have removed some of the products from purchase where it is not necessary.

Task-4972320


vava-odoo (approved):
@robodoo r+

PR #913, opened 2025-07-28
[FIX] beverage_distributor: add product and update quantities

dhrs-odoo:
In this PR we have added few products and also updated the inhand quantity.

Task-4972365


vava-odoo (approved):
@robodoo squash
@robodoo r+

PR #935, opened 2025-07-31
[FIX] automobile: fix automation / action perpetual loop

vava-odoo:
Before this commit, the automation would be triggered on every stock.move.line outgoing and done that is edited. The problem is that the linked server action writes on the stock.move.line, leading to an unlimited loop.

To get rid of this wrong behaviour, this commits adds a pre-domain on the base.automation such that it is only triggered on new stock.move.line meeting the filter domain. The edition done by the server action no more triggers the automation.


vava-odoo:
@robodoo r+

PR #957, opened 2025-08-06
[FIX] guest_house: fix confirming a sale order that caused an error

arthurrrm:
Confirming all the SO at once was causing a shift conflict in planning and prevented the module from installing.
This started to happen after a fix in Rental.

Now we split the confirmation of the SOs to treat first the ones getting picked-up/returned, this compute the potential conflicts before we confirms the other SOs, and solves the problem


tfr-odoo:
@robodoo r+

tfr-odoo:
@robodoo r+

PR #958, opened 2025-08-06
[ADD] industry_restaurant: added the supplier to products and appointment records

dhrs-odoo:
In this PR, the supplier "My Drink Vendor"  to the following products:
- Blond beer 
- Cheese 250gr 
- Ketchup 500ml
- Sanitizer 250ml

Additionally, the appointment records have been updated to display entries 
corresponding to the database generation date and the following day.

Task-4999831

### DIFF
--- a/industry_restaurant/demo/appointment_booking_line.xml
+++ b/industry_restaurant/demo/appointment_booking_line.xml
@@ -50,4 +50,54 @@
     <field name="capacity_reserved">4</field>
     <field name="calendar_event_id" ref="calendar_event_5"/>
   </record>
+  <record id="appointment_booking_line_11" model="appointment.booking.line">
+    <field name="appointment_resource_id" ref="appointment_resource_5"/>
+    <field name="capacity_reserved">4</field>
+    <field name="calendar_event_id" ref="calendar_event_11"/>
+  </record>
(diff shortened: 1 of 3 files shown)


pivi-odoo (changes requested):
Thank you for your patch.

Functionally it seems to be as intended in the task, so that part is fine.

I would just suggest a little improvement:
- Could you change the values of the `product_supplierinfo` lines, instead of most of them being (1,1), for a bit of variety / being more plausible.


PS: Not for this PR, but for future work: Know that industry modules are supposed to work on database that have *no demo* data installed from their principal module.
For example this module makes reference to data in `pos_restaurant` like the `coke` and others. This breaks the install of the module for fresh new databases with no demo data, and if in the future the standard XMLID changes by the R&D.

dhrs-odoo:
Hello @pivi-odoo 
I discussed this with @Jean-boje and he said it is intended to have 1$ price on it as he wants to demonstrate the suggestion feature but not particularly to have different prices on products

pivi-odoo (approved):
Ok, if the PO wanted that explicitly, it's ok then.

Therefor it's good to go IMHO.

Thank you for your work.

@tfr-odoo this PR is ok.

tfr-odoo:
@robodoo r+

PR #965, opened 2025-08-11
[FIX] *: fix static values in the website view

dhrs-odoo:
In this PR we have used reference instead of passing static values in the website view.

Task-4980201

### DIFF
--- a/architects/data/website_view.xml
+++ b/architects/data/website_view.xml
@@ -103,7 +103,7 @@
                                                                     <span class="s_website_form_label_content"/>
                                                                 </label>
                                                                 <div class="col-sm">
-                                                                    <input type="hidden" class="form-control s_website_form_input" name="user_id" value="2"/>
+                                                                    <input type="hidden" class="form-control s_website_form_input" name="user_id" t-att-value="request.env.ref('base.user_admin').id"/>
                                                                 </div>
                                                             </div>
                                                         </div>
--- a/certification_organism/data/website_view.xml
+++ b/certification_organism/data/website_view.xml
@@ -142,7 +142,7 @@
                                                                     <span class="s_website_form_label_content" />
                                                                 </label>
                                                                 <div class="col-sm">
-                                                                    <input type="hidden" class="form-control s_website_form_input" name="team_id" value="1" />
+                                                                    <input type="hidden" class="form-control s_website_form_input" name="team_id" t-att-value="request.env.ref('sales_team.team_sales_department').id" />
                                                                 </div>
                                                             </div>
                                                         </div>
--- a/corporate_gifts/data/website_view.xml
+++ b/corporate_gifts/data/website_view.xml
@@ -119,31 +119,31 @@
                                                                     <div class="row s_col_no_resize s_col_no_bgcolor s_website_form_multiple" data-name="tag_ids" data-display="horizontal">
                                                                         <div class="checkbox col-12 col-lg-4 col-md-6">
                                                                             <div class="form-check">
-                                                                                <input type="checkbox" class="s_website_form_input form-check-input" id="osinp1mz79g0" name="tag_ids" value="2" data-fill-with="undefined"/>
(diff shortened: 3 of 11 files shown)


fitness/data/appointment_view.xml
```diff
@@ -3,7 +3,7 @@
     <template id="custom_appointment_info_view" inherit_id="appointment.appointment_info">
         <xpath expr="//div[@t-if=&quot;not based_on_users and appointment_type.resource_manage_capacity&quot;]" position="replace">
             <div t-if="slots and max_capacity" class="mb-4">
-                <input type="hidden" name="resourceCapacity" value="1"/>
+                <input type="hidden" name="resourceCapacity" t-att-value="request.env.ref('fitness.appointment_type_1').id"/>
```
vava-odoo: This one is particular: we want to set 1. The goal is that a person can only subscribe for 1 person at a time (himself).
It should either be configured differently, or this particular case should get an exception in your test.

tests/test_generic/tests/test_xml.py
```diff
@@ -470,3 +471,36 @@ def _check_dates_are_relative(self, root, file_name):
...
+                tag.get("name") or tag.get("id") or "",
+                file_name,
+                line,
+            )
+
+        for tag in root.xpath("//a[@href]"):
+            href = tag.get("href")
+            if tag.get("t-att-href") or tag.get("t-attf-href"):
+                continue
+            if href and re.search(r"^/shop/[\w-]+-\d+$", href):
```
vava-odoo: Why only /shop? Could also be /event, /appointment, ... 
Can't you get something more generic?

tests/test_generic/tests/test_xml.py
```diff
@@ -470,3 +471,41 @@ def _check_dates_are_relative(self, root, file_name):
                         model_name,
                         file_name,
                     )
+
+    def _check_static_values_in_inputs(self, root, file_name):
+        allowed_inputs = {"resourceCapacity"}
```
abdrahmanrashed: Minor edit, feel free to ignore.
```suggestion
        ALLOWED_INPUTS = ["resourceCapacity"]
```
Use caps case to set variable to const in python.
If you have a small number of elements using list declaration instead of set is better the practice because it has less overhead.

tests/test_generic/tests/test_xml.py
```diff
@@ -470,3 +471,41 @@ def _check_dates_are_relative(self, root, file_name):
...
+        for tag in root.xpath("//input[@value] | //option[@value]"):
+            value = tag.get("value")
+            if not (value and value.isdigit()):
+                continue
```
abdrahmanrashed, line 480: why are we ignoring digits?

tests/test_generic/tests/test_xml.py
```diff
@@ -470,3 +471,41 @@ def _check_dates_are_relative(self, root, file_name):
...
+            if tag.get("name") in allowed_inputs:
+                continue
+
+            line = getattr(tag, "sourceline", "?")
+            _logger.warning(
+                "Static value '%s' found in <%s name='%s'> in %s (line %s). "
+                "Please use t-att-value=\"request.env.ref('module.record').id\" instead.",
+                value,
+                tag.tag,
+                tag.get("name") or tag.get("id") or "",
```
abdrahmanrashed:
```suggestion
            name = tag.get("name")
            if name in allowed_inputs:
                continue

            line = getattr(tag, "sourceline", "?")
            _logger.warning(
                "Static value '%s' found in <%s name='%s'> in %s (line %s). "
                "Please use t-att-value=\"request.env.ref('module.record').id\" instead.",
                value,
                tag.tag,
                name or tag.get("id") or "",
```
No need to retrieve name more than once.

tests/test_generic/tests/test_xml.py
dhrs-odoo: the goal is to catch hardcoded numeric IDs only, if we remove the .isdigit() then it will show warnings for perfectly fine values

tests/test_generic/tests/test_xml.py
dhrs-odoo: later on we might have to add more names so for that reason set is better

vava-odoo (approved):
Thanks both!
@robodoo r+

PR #988, opened 2025-08-22
[ADD] utils: utils to zip an industry and its dependencies

arthurrrm:
task-4797788

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+import ast
+import os
+import shutil
+from io import BytesIO
+from zipfile import ZipFile, ZIP_DEFLATED
+from odoo.tools import file_open, file_open_temporary_directory
+
+
+def get_zip(module_name, env):
+    output = BytesIO()
+    zf = ZipFile(output, "w", ZIP_DEFLATED)
+    external_modules = get_external_dependencies('industry/' + module_name, env)
+    dirs = ['industry/' + module_name] + ['industry/' + module for module in external_modules]
+    for dir in dirs:
+        for dirname, subdirs, files in os.walk(dir):
+            zip_dirname = os.sep.join(dirname.split(os.sep)[1:])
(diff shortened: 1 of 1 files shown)


utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+def get_external_dependencies(dir, env):
+    manifest_file = dir + '/__manifest__.py'
+    if not os.path.exists(manifest_file):
+        raise FileNotFoundError(f"Manifest file not found: {manifest_file}")
+    with open(manifest_file, encoding="utf-8") as f:
+        manifest = ast.literal_eval(f.read())
+    dependencies = set(manifest.get('depends', ['base']))
+    known_modules = env['ir.module.module'].search([]).mapped('name')
+    return dependencies.difference(known_modules)
```
vava-odoo:
```suggestion
```

utils.py
```diff
@@ -0,0 +1,31 @@
...
+from zipfile import ZipFile, ZIP_DEFLATED
+
+
+def get_zip(module_name, env):
+    output = BytesIO()
+    zf = ZipFile(output, "w", ZIP_DEFLATED)
+    external_modules = get_external_dependencies('industry/' + module_name, env)
+    dirs = ['industry/' + module_name] + ['industry/' + module for module in external_modules]
+    for dir in dirs:
+        for dirname, subdirs, files in os.walk(dir):
```
vava-odoo:
```suggestion
    for dirname, subdirs, files in os.walk('industry/' + module_name):
```

vava-odoo (changes requested):

utils.py
```diff
@@ -0,0 +1,31 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+import ast
+import os
+from io import BytesIO
+from zipfile import ZipFile, ZIP_DEFLATED
+
+
+def get_zip(module_name, env):
```
vava-odoo, line 7: `env` is not needed anymore, but I bet it's better to keep the same signature for the ease the work from runbot team

PR #1022, opened 2025-08-28
[FIX] theater: Impossible to log in

amas-odoo:
The field definition contained `eval="['website.template_header_sales_one']"/>` which is broken in standard. This caused loading errors when installing or upgrading the module.

Upstream Odoo task reference: https://www.odoo.com/odoo/project/974/tasks/5003249


vava-odoo (approved):
@robodoo r+

PR #1035, opened 2025-08-29
[REM] micro_brewery: remove useless view

arthurrrm:
Removed an inactive view that was not used anywhere.


vava-odoo (approved):
@robodoo r+

PR #1040, opened 2025-08-29
[FIX] gallery,micro_brewery: prevent error on opening product page

agbr-odoo:
Currently an error occurs when opening product page on website.

**Steps to replicate:**
* Install `micro_brewery` with demo data.
* Website > Shop > Open any product.

**Error**:
`ValueError: Element'<xpath expr='//div[@id='o_product_terms_and_share']'>' cannot be located in parent view`

**Root cause:**
After commit [1], the div at [2] was removed, but the demo data `XPath` expressions in the `gallery` and `micro_brewery` modules still reference it, causing error.

**Solution:**
* Modify the xpath template to match with the updated code.

[1]:
https://github.com/odoo/odoo/commit/bbb2d98d9ab97ce729d59b9858b63daccf5434e2#diff-635040c730ccbb8e4f6f2bf6472cf48a0cc9b8a3d488289dabdfce8096811915L2078

[2]:
https://github.com/odoo/odoo/blob/981a3f7afef6bf321b956b85e453ba85b83e43ed/addons/website_sale/views/templates.xml#L1884-L1887

sentry-6781422394

### DIFF
--- a/gallery/demo/website_view.xml
+++ b/gallery/demo/website_view.xml
@@ -422,17 +422,17 @@
   </record>
   <record id="ir_ui_view_3574" model="ir.ui.view">
     <field name="arch" type="xml">
-      <data inherit_id="website_sale.product" active="True" name="Share Buttons" priority="22">
-        <xpath expr="//div[@id='o_product_terms_and_share']" position="inside">
+      <data inherit_id="website_sale.product_terms_and_conditions" active="True" name="Share Buttons" priority="22">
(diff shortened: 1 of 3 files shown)


maan-odoo (approved):
LGTM :)

Also good to fix typo `Shar` here 
https://github.com/odoo/industry/blob/d6dd775b34191dfd6005a96844523f024fe9491b/gallery/demo/website_view.xml#L428

PR #1049, opened 2025-09-02
[REM] solar_installation: remove repair qty warning wizard from demo

lugr-odoo:
The insufficient quantity warning wizard is being removed from the repair module.

Task ID: [4890096](https://www.odoo.com/odoo/my-tasks/4890096)

See odoo/odoo#225095


vava-odoo (approved):
Thanks!
@robodoo r+

PR #1050, opened 2025-09-02
[IMP] excise_management: handle sales taxes for fiscal deposits

vava-odoo:
Before this commit, when selling a product with an excise tax to a partner with a fiscal position which is a fiscal deposit, the excise tax was added to the sale order line.

This commit improves the feature by adding a replacement tax to the non-fiscal deposit fiscal positions, such that the excise tax is not taking into account anymore. It also automatically handles new fiscal positions and new excise categories.

task-5030707


vava-odoo:
@robodoo r+

PR #1061, opened 2025-09-04
[FIX] concert_halls: fix end after begin with relativedelta

vava-odoo:
Before this commit, the use of relativedelta makes it possible to have a start date after the end date (if new was the end of the week). This commit makes it more robust.


vava-odoo:
@robodoo r+

PR #1065, opened 2025-09-04
[FIX] marketing_agency: disable header templates in website views

dhrs-odoo:
In this PR, we comment out the website header template views to work around a standard bug that breaks industry module views and causes related issues


event_management/demo/website_theme_apply.xml
```diff
@@ -49,6 +49,7 @@
     <value eval="{'active': False}" />
   </function>
 
+
```
vava-odoo: Useless change :upside_down_face:

PR #1082, opened 2025-09-09
[FIX] tests: allow changing description on PO line + allow reorder apps without studio

vava-odoo:


vava-odoo:
@robodoo override=ci/industry
@robodoo r+

PR #1084, opened 2025-09-09
[FIX] booking_engine: get the real rental unit price in booking engine

lvsz:
Versions
--------
- master

Steps
-----
1. Use the `campsite` industry;
2. go to /shop
3. book 2 nights on a large pitch for 2 people;
4. go to checkout & pay.

Issue
-----
Order couldn't be confirmed because the order amount changed post-payment.

Cause
-----
The booking engine automation action reset rental prices when a linked planning slot has its dates changed. It computes this rental price by getting the first `product.pricing` on the product template, ignoring any pricing differences for product variants.

Solution
--------
Update the rental dates to the slot dates, and use `_reset_price_unit` which gets the correct unit price.

opw-5075564


vava-odoo (approved):
@robodoo r+

PR #1089, opened 2025-09-10
[FIX] bakery: fix traceback during checkout extra step

dhrs-odoo:
Before this commit when we do the checkout extra step traceback occurs. In this commit we have fixed the traceback issue

Task-5075967


bakery/demo/website_view.xml
```diff
@@ -51,7 +51,7 @@
             </t>
         </t>
         </field>
-        <field name="key">website_sale.extra_info</field>
+        <field name="key">bakery.website_sale.extra_info</field>
```
vava-odoo:
```suggestion
        <field name="key">bakery.extra_info</field>
```

bakery/data/res_config_settings.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="res_config_settings_enable_bakery" model="res.config.settings">
+        <field name="enabled_extra_checkout_step" eval="1"/>
```
vava-odoo, line 4: Does it work without the setting (just by activating the view)?

dhrs-odoo: No

vava-odoo: Then we need to find another way for master :shrug:

PR #1108, opened 2025-09-12
[FIX] booking_engine,campsite: small UI fixes

amas-odoo:
- campsite: ensured bungalows are available to rent (no attribute if only 'no_variant' type).
- booking_engine: 
    - renamed breakfast attribute values to "Included" / "Not included".
    - colored schedule by partner_id (<attribute name="color">partner_id</attribute>).
    - Settings: added units option.
    - Set all room/stay offer products uom_id to "per unit".

Task ID: 5075526

### DIFF
--- a/booking_engine/data/ir_ui_view.xml
+++ b/booking_engine/data/ir_ui_view.xml
@@ -354,6 +354,7 @@
         <xpath expr="/gantt" position="attributes">
           <attribute name="form_view_id">%(booking_engine_schedule_form)d</attribute>
           <attribute name="multi_create_view"/>
+          <attribute name="color">partner_id</attribute>
         </xpath>
     </field>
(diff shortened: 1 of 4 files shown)


campsite/data/product_attribute.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+  <record id="booking_engine.product_attribute_11" model="product.attribute" forcecreate="1">
+    <field name="name">Air Conditioning</field>
+    <field name="create_variant">always</field>
```
vava-odoo: I don't get why we need to do that only for campsite. Seems like a workaround, but not a true fix. Isn't there something to do for product with only no_variant attribute to be displayed online?

vava-odoo (approved):
@robodoo r+ squash

PR #1133, opened 2025-09-16
[IMP] construction: add product cost and planning roles

amas-odoo:
- Set product cost as a fraction of the sale price.
- Added 1 material resource per existing role.
- Added new "Worker" role and assign.
- Added new "Work Manager" role and assign.

Task ID: 4236888

### DIFF
--- a/construction/data/product_product.xml
+++ b/construction/data/product_product.xml
@@ -51,155 +51,206 @@
 
     <record id="product_product_54" model="product.product">
         <field name="product_tmpl_id" ref="product_template_54"/>
+        <field name="standard_price" model="product.template" eval="0.52 * obj().env.ref('construction.product_template_54').list_price"/>
     </record>
     <record id="product_product_14" model="product.product">
         <field name="product_tmpl_id" ref="product_template_14"/>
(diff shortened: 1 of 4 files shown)


construction/data/product_product.xml
```diff
@@ -51,155 +51,206 @@
 
     <record id="product_product_54" model="product.product">
         <field name="product_tmpl_id" ref="product_template_54"/>
+        <field name="standard_price" model="product.template" eval="0.5 * obj().search([('id', '=', ref('product_template_54'))]).list_price"/>
```
vava-odoo:
```suggestion
        <field name="standard_price" model="product.template" eval="0.5 * obj().env.ref('product_template_54').list_price"/>
```

construction/data/product_product.xml
```diff
@@ -51,155 +51,206 @@
...
     </record>
     <record id="product_product_14" model="product.product">
         <field name="product_tmpl_id" ref="product_template_14"/>
+        <field name="standard_price" model="product.template" eval="0.5 * obj().search([('id', '=', ref('product_template_14'))]).list_price"/>
```
vava-odoo: Could you randomly assign different values instead of all 0.5, so that the db is more realistic?
A two digit number between 0.3 and 0.7:
0.34
0.56
0.61
...

PR #1135, opened 2025-09-16
[FIX] real_estate: correctly apply 4/3 thumbnail ratio on shop page

chga-odoo:
- The previous inheritance targeted section[@id='o_wsale_products_grid'], which conflicted with the default snippet logic and prevented the landscape (4/3) ratio from being applied.

- This fix adjusts the view ir_ui_view_thumb_4_3 to extend div[@id='o_wsale_container'] instead, ensuring that the class o_wsale_context_thumb_4_3 is properly added and the 4/3 thumbnail design is applied as expected.

Task ID: 4937043

### DIFF
--- a/real_estate/demo/website_view.xml
+++ b/real_estate/demo/website_view.xml
@@ -494,20 +494,7 @@
     <field name="type">qweb</field>
     <field name="website_id" ref="website.default_website"/>
   </record>
-  <!-- <record id="ir_ui_view_thumbnails_design" model="ir.ui.view">
-    <field name="name">Thumbnails Design</field>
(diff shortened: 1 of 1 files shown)


real_estate/demo/website_view.xml
```diff
@@ -494,19 +494,6 @@
     <field name="type">qweb</field>
     <field name="website_id" ref="website.default_website"/>
   </record>
-  <record id="ir_ui_view_thumbnails_design" model="ir.ui.view">
```
vava-odoo, line 497: I see the same in furniture_store. Shouldn't it be fixed there as well?

Issue #1136, opened 2025-09-16
it_hardware fails to install using odoo/18.0 and enterprise/18.0


mrlowe:
```
odoo.tools.convert.ParseError: while parsing /mnt/industry/it_hardware/data/helpdesk_config.xml:3, somewhere inside
<record id="helpdesk.helpdesk_team1" model="helpdesk.team" forcecreate="True">
        <field name="name">Customer Care</field>
        <field name="use_product_returns" eval="True"/>
        <field name="use_product_repairs" eval="True"/>
    </record>
```

Which appears to be caused by an error in loading helpdesk_repair:

```
2025-09-16 15:26:11,531 1 INFO odoo_db odoo.addons.base.models.ir_module: ALLOW access to module.button_immediate_install on ['Helpdesk Repair'] to user __system__ #1 via 172.25.0.1 
2025-09-16 15:26:11,532 1 INFO odoo_db odoo.addons.base.models.ir_module: User #1 triggered module installation 
2025-09-16 15:26:11,534 1 WARNING odoo_db odoo.modules.loading: Transient module states were reset 
2025-09-16 15:26:11,544 1 ERROR odoo_db odoo.modules.registry: Failed to load registry
```

Fix incoming.

PR #1137, opened 2025-09-16
[FIX] it_hardware: fails to install on odoo/18.0 and enterprise/18.0

mrlowe:
PR against 18.0 branch as this is where it was found, but probably present in 19.0 as well.

Description in #1136


mrlowe:
@vava-odoo thanks for the super easy change request! completed.

PR #1140, opened 2025-09-17
[FIX] gallery: publish products to avoid website errors

amas-odoo:
Fixed issue where unpublished products caused errors on the website.

Task ID: 5095343


vava-odoo (approved):
@robodoo r+ squash

PR #1157, opened 2025-09-23
[FIX] metal_fabricator: fix changes in metal fabricator

chga-odoo:
Added description field to the Sale Order Line.
Added project top bar
Adjusted project task stage so tasks now move to the correct stage in the Metal Fabricator flow.

Task ID: 5092711


vava-odoo (approved):
@robodoo r+ squash

PR #1172, opened 2025-09-25
[IMP] *: add live preview link and website link in appstore for all industry modules

vava-odoo:
**[IMP] *: add url for saas trial** 

This commit adds url to the manifest so that the "Live Preview" button on the appstore redirects to the saas trial form on odoo.com.

**[IMP] *: add website key in manifest**

This commit adds website to the manifest so that there is a link to the dedicated page on odoo.com.

task-4951169


tests/test_generic/tests/test_manifest.py
```diff
@@ -68,6 +70,12 @@ def _validate_manifest(self, module, manifest_data):
                     self.assertIn(value, CATEGORIES, f"Invalid category '{value}' not in {CATEGORIES}")
                 else:
                     self.assertNotIn(value, CATEGORIES, f"Module category '{value}' should not be an industry category: {CATEGORIES}")
+            elif key == 'url' and module in self.installed_industries:
+                self.assertEqual(value, f"https://www.odoo.com/trial?industry&selected_app={module}", f"The url should link to the trial page: https://www.odoo.com/trial?industry&selected_app={module}")
+            elif key == 'website' and module in self.installed_industries:
+                self.assertTrue(
+                    value.startswith("https://www.odoo.com/industries/") or value == "https://www.odoo.com/all-industries", "The url should link to the odoo.com page."
+                )
             elif key in ['data', 'demo']:
```
abdrahmanrashed:
```suggestion
            if module in self.installed_industries:
                if key == 'url':
                    self.assertEqual(value, f"https://www.odoo.com/trial?industry&selected_app={module}", f"The url should link to the trial page: https://www.odoo.com/trial?industry&selected_app={module}")
                elif key == 'website':
                    self.assertTrue(
                        value.startswith("https://www.odoo.com/industries/") or value == "https://www.odoo.com/all-industries", "The url should link to the odoo.com page."
                    )
                elif key == 'category':
                    self.assertIn(value, CATEGORIES, f"Invalid category '{value}' not in {CATEGORIES}")
                else:
                    self.assertNotIn(value, CATEGORIES, f"Module category '{value}' should not be an industry category: {CATEGORIES}")
            if key in ['data', 'demo']:
```

abdrahmanrashed (commented):
Amazing work as always.

Only one small comment, no need to iterate through the installed_industries more than once.

tests/test_generic/tests/test_manifest.py
vava-odoo: I pushed your suggestion, but feel like we lost clarity. I wonder if performance is an issue here, because it's only checking if a key is in a (small) set.

vava-odoo:
@robodoo r-

tests/test_generic/tests/test_manifest.py
```diff
@@ -63,13 +65,19 @@ def _validate_manifest(self, module, manifest_data):
             self.assertIsInstance(value, expected_type, f"Wrong type for '{key}', expected {expected_type}")
             if expected_value:
                 self.assertEqual(value, expected_value, f"Wrong {key} '{value}' in manifest, it should be {expected_value}")
-            if key == 'category':
-                if module in self.installed_industries:
-                    self.assertIn(value, CATEGORIES, f"Invalid category '{value}' not in {CATEGORIES}")
-                else:
-                    self.assertNotIn(value, CATEGORIES, f"Module category '{value}' should not be an industry category: {CATEGORIES}")
             elif key in ['data', 'demo']:
```
Pierre-savp: this might move to row 80 or be done as a separate if no?

vava-odoo: well spotted! First option doesn't work because it should apply to all. But definitely a if on its own

PR #1179, opened 2025-09-26
[FIX] * : add required email_from value in 'mailing.mailing'

maan-odoo:
*= dropshipping, members_club, student_organization

Users are currently unable to install the above industries because the required `email_from` field is missing in the mailing data.

A recent commit [1] introduced a constraint that raises an error if the `email_from` field is not set and `mailing_type` is mail (default value).

This commit will fix the above issue by adding the `email_from` value in the data.

[1] - https://github.com/odoo/odoo/commit/92c5b0699dffab98f55ed1a54a0236dbc4b5626a

Task-No

### DIFF
--- a/dropshipping/data/mailing_mailing.xml
+++ b/dropshipping/data/mailing_mailing.xml
@@ -587,6 +587,7 @@ width:75%;
 </table>
 ]]></field>
     <field name="medium_id" ref="utm.utm_medium_email"/>
+    <field name="email_from">dropshipping@example.com</field>
     <field name="user_id" ref="base.user_admin"/>
     <field name="reply_to_mode">new</field>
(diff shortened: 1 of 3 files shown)


dropshipping/data/mailing_mailing.xml
```diff
@@ -587,6 +587,7 @@ width:75%;
 </table>
 ]]></field>
     <field name="medium_id" ref="utm.utm_medium_email"/>
+    <field name="email_from">dropshipping@example.com</field>
```
vava-odoo: Thanks for the fix. But couldn't we use the mail address of the admin? It would make more sense than a random mail.

maan-odoo: Hey @vava-odoo 

I have updated it; now it will set the admin email as `email_from` if it is available, otherwise a random mail.

vava-odoo:
@robodoo r+

PR #1194, opened 2025-10-03
[IMP] utils: industry run script improvements.

abdrahmanrashed:
- swapped db dump logic for dp duplicate for increased preformance.
- set [-i] to be optional and added -n for db name input

### DIFF
--- a/run_industry.sh
+++ b/run_industry.sh
@@ -3,8 +3,9 @@ set -euo pipefail
 
 # Usage function
 usage() {
-  echo "Usage: ./industry/run_industry.sh -i <industry-name> [-d] [-t] [-r | -h]"
-  echo "  -i <industry-name>   (re)Install this industry"
+  echo "Usage: ./industry/run_industry.sh -n <industry-name> [-i] [-d] [-t] [-r | -h]"
+  echo "  -n <industry-name>   Name of the industry to run"
+  echo "  -i                   Import this industry"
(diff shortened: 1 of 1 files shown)


run_industry.sh
```diff
@@ -3,8 +3,9 @@ set -euo pipefail
 
 # Usage function
 usage() {
-  echo "Usage: ./industry/run_industry.sh -i <industry-name> [-d] [-t] [-r | -h]"
-  echo "  -i <industry-name>   (re)Install this industry"
+  echo "Usage: ./industry/run_industry.sh -n <industry-name> [-i] [-d] [-t] [-r | -h]"
+  echo "  -n <industry-name>   Name of the industry to run"
+  echo "  -i                   (re)Install this industry"
```
vava-odoo: Wording is misleading here. What about Import?
```suggestion
  echo "  -i                   Import this industry"
```

utils.py
```diff
@@ -23,7 +23,7 @@ def get_zip(module_name, env):
 def get_dependencies(dir):
     manifest_file = dir + '/__manifest__.py'
     if not os.path.exists(manifest_file):
-        raise FileNotFoundError(f"Manifest file not found: {manifest_file}")
+        return set()
```
vava-odoo: Why wouldn't you raise here?

vava-odoo (approved):
@robodoo r+

PR #1609, opened 2026-01-27
[FIX] booking_engine: fix work schedule and product time adjustment

abdrahmanrashed:
- set Rental 24/7 working schedule to actually be 7 days week with 24h per day
- sync SOL date description with SO rental time
- temporarily remove SO timezone shift to properly set rental period adjustment
- added images for `booking_engine` and `deposit_management`

task-5794820


vava-odoo:
@abdrahmanrashed this fix is probably useless now, but maybe we could add the images

abdrahmanrashed:
> @abdrahmanrashed this fix is probably useless now, but maybe we could add the images

@vava-odoo it is only removed 19.2+, we should probably have it for 19.0 and 19.1 and push only the images for 19.2+

vava-odoo:
@Pierre-savp

Pierre-savp:
@vava-odoo @abdrahmanrashed 
Agreed that with the removal of time constrain, some lines here might be dropped / simplified
There already is a ticket to remove the time constrain before 19.2 so in the end this might only contain pic update

vava-odoo:
@robodoo fw=no r+

PR #1633, opened 2026-02-04
[IMP] *: Do not pollute message and notification box when installing an industry

amas-odoo:
- Remove res_users.xml forcing notification_type='inbox' so modules generate
mails again.
- Read generated mails via _read_group and browse related records.
- Inspect linked mail.activity and handle upsell activities separately:
   - Identify upsell activities from note content.
   - Ignore upsell activities without external IDs when their count matches
     generated mails.
   - Process remaining activities to detect missing contexts.
- Suggest required contexts to prevent message/mail pollution.
- Log warnings with external IDs and recommended contexts.
- Add validation on XML <function> nodes to ensure required mail-prevention
contexts are set.
- Introduce helper log_warning to standardize missing-context warnings.
- Add specific checks:
   - slide.channel: ensure publish_template_id is set to False.
   - sale.order.action_confirm: require mail_auto_subscribe_no_notify and
      mail_notrack when subscription SO is present.
   - Functions defined in FUNCTION_CONTEXT_DICT_WITHOUT_USER:
      enforce mail_auto_subscribe_no_notify.
    - write calls updating user_id / user_ids: require
      mail_auto_subscribe_no_notify to prevent unwanted followers/mails.
- Safely parse function context using literal_eval with fallback handling.

`Context Used`

- mail_auto_subscribe_no_notify: Prevents notifications that are normally sent
when users are automatically added as followers. By default this is False,
meaning notifications are sent — enable it to stay silent.
- mail_activity_quick_update: When creating activities for other users, this
avoids sending them a notification about the new activity.
- mail_notrack: During create or write, disables field change tracking so no
automatic “value changed” messages are posted.
- no_mail_to_attendees: Stops email notifications from being sent to event attendees.

Task ID: 5262306

### DIFF
--- a/3pl_logistic_company/demo/crm_lead.xml
+++ b/3pl_logistic_company/demo/crm_lead.xml
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='UTF-8'?>
-<odoo noupdate="1">
+<odoo noupdate="1" context="{'mail_auto_subscribe_no_notify': True}">
   <record id="crm_lead_1" model="crm.lead">
     <field name="name">Contact with Auto 5</field>
     <field name="user_id" ref="base.user_admin"/>
--- a/tests/test_generic/tests/test_industry_mail_behavior.py
+++ b/tests/test_generic/tests/test_industry_mail_behavior.py
@@ -2,28 +2,91 @@
 
 import logging
 
+
 from odoo.tests import tagged
 from .industry_case import IndustryCase
 
 _logger = logging.getLogger(__name__)
 
 
+RECORD_CONTEXT_MODELS_DICT = {
+    'calendar.event': ['no_mail_to_attendees', 'mail_notrack'],
+    'crm.lead': ['mail_auto_subscribe_no_notify'],
+    'event.event': ['mail_auto_subscribe_no_notify'],
+    'helpdesk.ticket': ['mail_notrack', 'mail_auto_subscribe_no_notify'],
+    'hr.applicant': ['mail_notrack'],
+    'hr.job': ['mail_auto_subscribe_no_notify'],
+    'mail.activity': ['mail_activity_quick_update'],
+    'mailing.mailing': ['mail_auto_subscribe_no_notify'],
+    'maintenance.request': ['mail_auto_subscribe_no_notify', 'mail_activity_quick_update'],
+    'mrp.eco': ['mail_auto_subscribe_no_notify'],
+    'pos.session': ['mail_auto_subscribe_no_notify'],
+    'project.project': ['mail_auto_subscribe_no_notify'],
+    'project.task': ['mail_auto_subscribe_no_notify'],
+    'purchase.order': ['mail_auto_subscribe_no_notify'],
+    'quality.check': ['mail_auto_subscribe_no_notify'],
+    'res.partner': ['mail_auto_subscribe_no_notify'],
+    'sale.order': ['mail_auto_subscribe_no_notify'],
+    'slide.channel': ['mail_auto_subscribe_no_notify'],
+    'stock.picking': ['mail_auto_subscribe_no_notify'],
+    'survey.survey': ['mail_auto_subscribe_no_notify'],
+}
+
+EXCEPTION_MODELS = [
+    'hr.expense',
+    'hr.leave',
+]
+
+
 @tagged('post_install', '-at_install')
 class IndustryMailBehaviorTestCase(IndustryCase):
-    def test_user_notification_type(self):
-        if not self.env['ir.module.module'].search_count([('demo', '=', True)], limit=1):
-            return
-        for user in self.env['res.users'].search([('group_ids', 'in', [self.env.ref('base.group_user').id, self.env.ref('base.group_system').id]), ('notification_type', '!=', 'inbox')]):
-            _logger.warning("Notification type should remain 'inbox' for all users, not the case for %s (%d).", user.name, user.id)
 
-    def test_mail_generated(self):
-        if self.env['ir.module.module'].search_count([('demo', '=', True)], limit=1):
-            return
+    def test_generic_mail_message_generated(self):
+        mails = self.env['mail.mail']._read_group([('model', '!=', False), ('model', 'not in', EXCEPTION_MODELS), ('state', '!=', 'outgoing')], ['res_id', 'model'])
 
-        mails = self.env['mail.mail'].search([('model', '!=', False), ('model', '!=', 'hr.expense')])
-        if mails:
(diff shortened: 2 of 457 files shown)


tests/test_generic/tests/test_industry_mail_behavior.py
```diff
@@ -2,30 +2,89 @@
...
+        for res_id, res_model, count in mails:
+            record = self.env[res_model].browse(res_id)
+            if not record:
+                continue
+            all_activities = getattr(record, 'activity_ids', False) or self.env['mail.activity']
+            upsell_activity = all_activities.filtered(lambda a: a.note and 'upsell' in a.note.lower())
+            upsell_activity_external_ids = upsell_activity._get_external_ids()
+            upsell_activity_without_external_ids = upsell_activity.filtered(lambda a: not upsell_activity_external_ids.get(a.id))
+            if len(upsell_activity_without_external_ids) == count:
+                continue
```
vava-odoo: Could you elaborate here? If I remember correctly, upsell activity type was removed a few versions ago
And I see only 2 industry with "upsell" mail.activity summary, but the type is a todo

tests/test_generic/tests/test_industry_mail_behavior.py
```diff
@@ -2,30 +2,89 @@
...
+
+            ext_ids = record._get_external_ids()
+            if not any(ext_ids.values()):
+                continue
+            for ext_id in next(iter(ext_ids.values())):
+                ctxs = RECORD_CONTEXT_MODELS_DICT.get(res_model)
+                if ctxs:
+                    result.setdefault(ext_id, set()).update(ctxs)
```
vava-odoo: the problem here is that if the model is not in RECORD_CONTEXT_MODELS_DICT but generates a mail, it won't be added to result, and therefore won't generate any warning. 
Wouldn't it be best to add ext_id to result right away, and then add the context keys later?
```suggestion
                result.setdefault(ext_id, set())
                if (ctxs := RECORD_CONTEXT_MODELS_DICT.get(res_model)):
                    result.update(ctxs)
```
Maybe we would need to adapt the warning in case result item value is empty (means this case is not yet covered)

tests/test_generic/tests/test_xml.py
```diff
@@ -94,20 +94,9 @@
...
 ]
 
-SKIP_CONTEXT_DICT_FOR_DEMO = {
-    'crm.lead': 'mail_auto_subscribe_no_notify',
-    'event.event': 'mail_auto_subscribe_no_notify',
-    'hr.job': 'mail_auto_subscribe_no_notify',
-    'mailing.mailing': 'mail_auto_subscribe_no_notify',
-    'project.project': 'mail_auto_subscribe_no_notify',
-    'project.task': 'mail_auto_subscribe_no_notify',
-    'stock.picking': 'skip_sms',
```
vava-odoo, line 104: why is skip_sms drop here?

tests/test_generic/tests/test_xml.py
```diff
@@ -593,32 +582,72 @@ def _check_static_values_in_inputs(self, root, file_name):
...
-            context_dict = literal_eval(record_context) if record_context else {}
-            if expected_context not in context_dict:
+            field_publish_template_id = record.xpath("./field[@name='publish_template_id']")
+            warn = False
+            if not field_publish_template_id:
+                warn = True
+            else:
+                for field in record.xpath("./field[@name='publish_template_id']"):
+                    publish_template_ref = field.get('ref') or field.get('eval')
+                    if publish_template_ref != 'False':
```
vava-odoo: Do we really want to force remove the linked template? Isn't there another way to skip the mail from demo?
We probably would like to set up a slide.channel cleanly, including the template.
What do you think?

3pl_logistic_company/demo/crm_lead.xml
```diff
@@ -1,14 +1,14 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo noupdate="1">
-  <record id="crm_lead_1" model="crm.lead">
+  <record id="crm_lead_1" model="crm.lead" context="{'mail_auto_subscribe_no_notify': True}">
```
vava-odoo: I think it would be cleaner to move the context to the parent `<odoo>` so that it defines only once per file

vava-odoo (commented):
Great work. I focused on the testing code for now. And I mostly have questions at this stage.

tests/test_generic/tests/test_industry_mail_behavior.py
amas-odoo: Yes, you are right, and you are also correct that the upsell activity is now of type To-Do. However, upsell activities are still generated with the To-Do type and identified by the “upsell” summary, and for these activities, email notifications are not suppressible.

You can also check the carpenter industry module in mail.notification for reference. 
Note: Email is not generated because we set notification type inbox

image

tests/test_generic/tests/test_xml.py
amas-odoo: While reviewing the industry modules, I noticed that emails are not being generated even without skip_sms, which is why I decided to omit it here. If you think it is still required, we can add it back.

3pl_logistic_company/demo/crm_lead.xml
amas-odoo: I assumed this was not available in 19.0. I will proceed with the necessary changes.

tests/test_generic/tests/test_xml.py
amas-odoo: I looked for a way to prevent mails for this but wasn’t able to find any. This is the only approach I came across. If there is another way to prevent the mails, I can adapt accordingly.

tests/test_generic/tests/test_xml.py
vava-odoo: maybe we could at some point investigate about sms, but let's remove for now. Maybe keep the context key where it is currently, we just won't add it in the near future. It's ok

tests/test_generic/tests/test_industry_mail_behavior.py
vava-odoo: Maybe we could then edit the summary? 
Could you point to the code checking the summary part of the activity?

tests/test_generic/tests/test_xml.py
vava-odoo: I don't see one right now. 
Maybe, what could be done, is set it to False at the beginning, add contents, followers, ... and then add the template afterwards. 
And in the testing, you ensure this is the last step. 
Would that work? Or is it overkilled?

vava-odoo (approved):
@robodoo r+

PR #1761, opened 2026-02-26
[IMP] real_estate: rely on standard prevent_sale instead of custom CTA override

chga-odoo:
The custom add-to-cart button override was removed and replaced by the standard website prevent_sale configuration for specific public product categories.

task - 5943972

### DIFF
--- a/real_estate/__manifest__.py
+++ b/real_estate/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Real Estate Agency',
-    'version': '1.5',
+    'version': '1.6',
     'category': 'Services',
     'author': 'Odoo S.A.',
     'depends': [
@@ -59,6 +59,7 @@
         'data/website_menu.xml',
         'data/website_theme_apply.xml',
         'data/uninstall_hook.xml',
+        'data/website.xml',
     ],
     'demo': [
         'demo/product_attribute_value.xml',
--- a/real_estate/data/website_view.xml
+++ b/real_estate/data/website_view.xml
@@ -100,7 +100,7 @@
                                 <span class="s_website_form_label_content">Subject</span>
                                 <span class="s_website_form_mark">          *</span>
                               </label>
-                              <input class="form-control s_website_form_input" type="text" name="name" required="" placeholder="" id="oy7pg7e0vg6c" data-fill-with="undefined"/>
+                              <input class="form-control s_website_form_input" type="text" name="name" id="oy7pg7e0vg6c" data-fill-with="undefined"/>
                             </div>
                             <div data-name="Field" class="s_website_form_field mb-3 col-12 s_website_form_required" data-type="text" data-translated-name="Field">
                               <label class="s_website_form_label" style="width: 250px;" for="o5mtutqkqd9a">
@@ -182,7 +182,7 @@
                             <span class="s_website_form_label_content">Your Name</span>
                             <span class="s_website_form_mark"> *</span>
                           </label>
-                          <input class="form-control s_website_form_input" type="text" name="name" required="" placeholder="" id="opv6ikrbw8i" data-fill-with="name"/>
+                          <input class="form-control s_website_form_input" type="text" name="name" id="opv6ikrbw8i" data-fill-with="name"/>
                         </div>
                         <div data-name="Field" class="s_website_form_field mb-3 col-12 s_website_form_required" data-type="email" data-translated-name="Field">
                           <label class="s_website_form_label" style="width: 500px;" for="o36rnwjacmkt">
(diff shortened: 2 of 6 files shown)


real_estate/demo/website.xml
```diff
@@ -13,5 +13,8 @@
         <field name="shop_gap">13px</field>
         <field name="product_page_image_layout">grid</field>
         <field name="product_page_image_width">66_pc</field>
+        <field name="prevent_sale">True</field>
+        <field name="prevent_sale_for">specific_categories</field>
+        <field name="prevent_sale_for_categories" eval="[(4, ref('real_estate.product_public_category_1')), (4, ref('real_estate.product_public_category_2'))]"/>
```
vava-odoo: this needs to be in data/

real_estate/data/website_view.xml
vava-odoo: Isn't this useless as well, now?

real_estate/__manifest__.py
vava-odoo: bump :-)

real_estate/data/website_view.xml
```diff
@@ -288,124 +288,115 @@
...
+                              <span class="s_website_form_mark">        *</span>
+                            </label>
+                            <input class="form-control s_website_form_input" type="text" name="partner_name" placeholder="" id="otpj9zeau4p" data-fill-with="commercial_company_name"/>
+                          </div>
+                          <div data-name="Field" class="s_website_form_field mb-3 col-12 s_website_form_model_required" data-type="char" data-translated-name="Field">
+                            <label class="s_website_form_label" style="width: 200px" for="otw2ya2767a">
+                              <span class="s_website_form_label_content">Subject</span>
+                              <span class="s_website_form_mark">    *</span>
+                            </label>
+                            <input class="form-control s_website_form_input" type="text" name="name" required="" placeholder="" id="otw2ya2767a" data-fill-with="name"/>
```
pgu-odoo:
```suggestion
                            <input class="form-control s_website_form_input" type="text" name="name" id="otw2ya2767a" data-fill-with="name"/>
```

real_estate/data/website_view.xml
```diff
@@ -288,124 +288,115 @@
...
+                          </div>
+                          <div hidden="true" data-name="Field" class="s_website_form_field mb-3 col-12" data-type="char" data-translated-name="Field">
```
pgu-odoo:
```suggestion
                          <div class="d-none" data-name="Field" class="s_website_form_field mb-3 col-12" data-type="char" data-translated-name="Field">
```

real_estate/data/website_view.xml
```diff
@@ -288,124 +288,115 @@
...
+  <record id="information_request" model="ir.ui.view">
+    <field name="name">Information Request</field>
+    <field name="key">real_estate.information_request</field>
+    <field name="active" eval="True"/>
+    <field name="type">qweb</field>
+    <field name="website_id" ref="website.default_website"/>
+    <field name="arch" type="xml">
+      <t name="Information Request" t-name="website.information_request">
+        <t t-call="website.layout">
+          <t t-set="logged_partner" t-value="request.env['website.visitor']._get_visitor_from_request().partner_id"/>
```
pgu-odoo: is it used anywhere?

pgu-odoo (commented):
@vava-odoo PR is ready for your review.
@chga-odoo Good work! A few comments are added.

real_estate/data/website_view.xml
vava-odoo: don't we need to use the new syntax of t-call since 19.2? Something like

```suggestion
        <t t-call="website.layout" logged_partner.f="request.env['website.visitor']._get_visitor_from_request().partner_id"

```

real_estate/i18n/real_estate.pot
vava-odoo, line 1: you can remove the changes to this file

real_estate/__manifest__.py
```diff
@@ -58,14 +57,14 @@
         'data/website_page.xml',
         'data/website_menu.xml',
         'data/website_theme_apply.xml',
+        'data/website.xml',
```
vava-odoo, line 62: requires a version update

vava-odoo (commented):
Well done. Just some small comments + needs to resolve conflicts. 
Thanks!

real_estate/data/website_view.xml
chga-odoo: d-none is not working here because of its override by class s_website_form_field probably, we can use 's_website_form_field_hidden' its working fine :+1:

real_estate/data/website_view.xml
```diff
@@ -288,124 +288,114 @@
...
+  <record id="information_request" model="ir.ui.view">
+    <field name="name">Information Request</field>
+    <field name="key">real_estate.information_request</field>
+    <field name="active" eval="True"/>
+    <field name="type">qweb</field>
+    <field name="website_id" ref="website.default_website"/>
+    <field name="arch" type="xml">
+      <t name="Information Request" t-name="website.information_request">
+        <t t-call="website.layout" logged_partner="request.env['website.visitor']._get_visitor_from_request().partner_id">
+          <t t-set="contactus_form_values" t-value="{ 'name': request.params.get('subject', ''),             'x_interested_in_id':request.params.get('product_id'),             }"/>
```
vava-odoo, line 316: why you didn't adapt this one as well?
I must admit I'm a bit lost, since this version seems to work as is...
@amas-odoo since you worked on this, could you share light?

amas-odoo: @vava-odoo 
The reason this approach wasn’t adapt is because variables passed as attributes in <t t-call="..."> are only accessible inside the called template, not in the current template.

In this case, contactus_form_values is required within the current template itself. If we try to pass contactus_form_values as an attribute, it won’t be accessible where it’s needed and can lead to errors.

That’s why we define it using <t t-set="contactus_form_values" ...> directly in the current template instead of passing it through t-call.

vava-odoo (approved):
@robodoo r+

PR #1855, opened 2026-03-16
[IMP] construction_developer: versatile work items pricing

frva-odoo:
Improves the work item model with the possibility to manage prices manually.
Before this PR, every work item had their cost and price defined as the sum of their components costs and prices.
Now, this works as before but we can check the fixed checkbox next to the margin to enable an alternative behavior:
Input manually a fixed margin or a fixed price. This margin is kept and impacts the price when the cost of components changes.

A second commit in this pr adds some tests for the construction module: 
- Test the creation of a work item related to a sale order, see if the product is created behind.
- Test that there is a single work item template at the same time for a product
- Test the cost, price and margin of a work item whenever components changes or work item values are changed directly

Task-5262058

### DIFF
--- a/construction_developer/data/base_automation.xml
+++ b/construction_developer/data/base_automation.xml
@@ -63,6 +63,32 @@
         <field name="name">Construction: Work Item - On save and unit cost/price change</field>
         <field name="trigger_field_ids" eval="[(6, 0, [ref('x_unit_cost_field_x_work_item'), ref('x_unit_price_field_x_work_item')])]"/>
     </record>
+    <record id="base_automation_product_product_on_write_unit_cost_price_change" model="base.automation">
+        <field name="model_id" ref="product.model_product_product" />
+        <field name="action_server_ids"
+            eval="[(6, 0, [ref('server_action_update_related_work_item_lines')])]" />
+        <field name="trigger">on_create_or_write</field>
+        <field name="name">Construction: Work Item - On save and cost/price change</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('product.field_product_product__lst_price'), ref('product.field_product_product__standard_price')])]"/>
+    </record>
+    <record id="base_automation_work_item_on_write_unit_price_when_margin_fixed" model="base.automation">
+        <field name="model_id" ref="x_work_item_model" />
+        <field name="action_server_ids"
+            eval="[(6, 0, [ref('server_action_update_margin_if_margin_fixed')])]" />
+        <field name="trigger">on_create_or_write</field>
+        <field name="filter_domain">[("x_is_margin_fixed", "=", True)]</field>
+        <field name="name">Construction: Work Item - On save and unit price change</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('x_unit_price_field_x_work_item')])]"/>
+    </record>
+    <record id="base_automation_work_item_on_write_unit_margin_or_cost_when_margin_fixed" model="base.automation">
+        <field name="model_id" ref="x_work_item_model" />
+        <field name="action_server_ids"
+            eval="[(6, 0, [ref('server_action_update_price_if_margin_fixed')])]" />
+        <field name="trigger">on_create_or_write</field>
+        <field name="filter_domain">[("x_is_margin_fixed", "=", True)]</field>
+        <field name="name">Construction: Work Item - On save and unit margin or cost change</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('x_unit_margin_field_x_work_item'), ref('x_unit_cost_field_x_work_item')])]"/>
+    </record>
     <record id="base_automation_work_item_on_write" model="base.automation">
         <field name="model_id" ref="x_work_item_model" />
         <field name="action_server_ids"
@@ -86,6 +112,15 @@
         <field name="trigger">on_create_or_write</field>
         <field name="filter_domain">[('x_is_template', '!=', False), ('x_product_id', '=', False)]</field>
         <field name="name">Construction: Work Item - On save of work item template without product</field>
+    </record>
+    <record id="base_automation_on_write_wi_margin_fixed_to_false" model="base.automation">
+        <field name="model_id" ref="x_work_item_model" />
+        <field name="action_server_ids"
+            eval="[(6, 0, [ref('server_action_recompute_price_with_components_prices')])]" />
+        <field name="trigger">on_create_or_write</field>
+        <field name="name">Construction: Work Item - On Margin Fixed write to False</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('x_is_margin_fixed_field_x_work_item')])]"/>
+        <field name="filter_domain">[('x_is_margin_fixed', '=', False)]</field>
     </record>
     <!-- 
     this is needed to set a default stage based on the project's
--- a/construction_developer/data/ir_actions_server.xml
+++ b/construction_developer/data/ir_actions_server.xml
@@ -124,6 +124,16 @@ record.write({'x_name': record.x_product_id.name})
 record.write({'x_unit_price': record.x_product_id.list_price, 'x_unit_cost': record.x_product_id.standard_price})
 ]]></field>
     </record>
+    <record id="server_action_update_related_work_item_lines" model="ir.actions.server">
+        <field name="name">Construction: Work Item - Update related wi lines on product cost/price change</field>
+        <field name="model_id" ref="product.model_product_product"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+wil = env['x_work_item_line'].search([('x_product_id', 'in', records.ids), ('x_work_item_id.x_is_template', '=', True)])
+for product in records:
+    wil.filtered(lambda line: line.x_product_id == product).write({'x_unit_price': product.list_price, 'x_unit_cost': product.standard_price})
+]]></field>
+    </record>
     <record id="server_action_update_product_cost" model="ir.actions.server">
         <field name="name">Construction: Work Item - Update product cost</field>
         <field name="model_id" ref="x_work_item_model"/>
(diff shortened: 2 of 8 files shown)


construction_developer/data/base_automation.xml
```diff
@@ -55,6 +55,32 @@
         <field name="name">Construction: Work Item - On save and unit cost/price change</field>
         <field name="trigger_field_ids" eval="[(6, 0, [ref('x_unit_cost_field_x_work_item'), ref('x_unit_price_field_x_work_item')])]"/>
     </record>
+    <record id="base_automation_product_product_on_write_unit_cost_price_change" model="base.automation">
+        <field name="model_id" ref="product.model_product_product" />
+        <field name="action_server_ids"
+            eval="[(6, 0, [ref('server_action_update_related_work_item_lines')])]" />
+        <field name="trigger">on_write</field>
```
vava-odoo: why on_write?

construction_developer/data/ir_actions_server.xml
```diff
@@ -113,6 +113,15 @@ record.write({'x_name': record.x_product_id.name})
...
 record.write({'x_unit_price': record.x_product_id.list_price, 'x_unit_cost': record.x_product_id.standard_price})
+]]></field>
+    </record>
+    <record id="server_action_update_related_work_item_lines" model="ir.actions.server">
+        <field name="name">Construction: Work Item - Update related wi lines on product cost/price change</field>
+        <field name="model_id" ref="product.model_product_product"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for r in records:
+    env['x_work_item_line'].search([('x_product_id', '=', r.id), ('x_work_item_id.x_is_template', '=', True)]).write({'x_unit_price': r.list_price, 'x_unit_cost': r.standard_price})
```
vava-odoo: probably better to search all template then filter in python. What do you think?
```suggestion
wil = env['x_work_item_line'].search([('x_product_id', 'in', records.ids), ('x_work_item_id.x_is_template', '=', True)])
for product in records:
    wil.filtered(lambda line: line.x_product_id == product).write({'x_unit_price': product.list_price, 'x_unit_cost': product.standard_price})
```

construction_developer/data/ir_actions_server.xml
```diff
@@ -135,6 +144,30 @@ record.write({'x_unit_price': record.x_product_id.list_price, 'x_unit_cost': rec
         <field name="evaluation_type">equation</field>
         <field name="value"><![CDATA[record.x_unit_price]]></field>
     </record>
+    <record id="server_action_update_margin_if_margin_fixed" model="ir.actions.server">
+        <field name="name">Construction: Work Item - Update margin on price change if margin is fixed</field>
+        <field name="model_id" ref="x_work_item_model" />
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+record['x_unit_margin'] = (record.x_unit_price - record.x_unit_cost) / record.x_unit_cost if record.x_unit_cost > 0 else False
```
vava-odoo: Don't you need to loop over records? Same for all below.
I know we always do it from the form view, but there are other ways.
Same remark for next actions as well.

construction_developer/data/ir_actions_server.xml
```diff
@@ -135,6 +144,30 @@ record.write({'x_unit_price': record.x_product_id.list_price, 'x_unit_cost': rec
...
+]]></field>
+    </record>
+    <record id="server_action_update_price_if_margin_fixed" model="ir.actions.server">
+        <field name="name">Construction: Work Item - Update price on price change if margin is fixed</field>
```
vava-odoo:
```suggestion
        <field name="name">Construction: Work Item - Update price on margin change if margin is fixed</field>
```

construction_developer/data/ir_ui_view.xml
```diff
@@ -150,16 +150,22 @@
...
+                        <label for="x_unit_price"/>
+                        <div class="o_row">
+                            <field name="x_unit_price" class="oe_inline mw-50" widget="monetary" force_save="1" readonly="not x_is_margin_fixed"/>
+                            <button type="action" class="btn btn-secondary oe_inline" name="%(server_action_recompute_price_with_components_prices)d" invisible="not x_is_margin_fixed" string="Recompute" help="Recomputes the unit price based on components"/>
+                        </div>
+                        <label for="x_unit_margin"/>
+                        <div class="o_row">
+                            <field name="x_unit_margin" widget="percentage" class="oe_inline mw-50" force_save="1" readonly="not x_is_margin_fixed"/>
+                            <label for="x_is_margin_fixed"/>
+                            <field name="x_is_margin_fixed" class="oe_inline" help="When set, unit price is computed based on unit margin"/>
```
vava-odoo: maybe force_save as well?

construction_developer/data/ir_ui_view.xml
```diff
@@ -150,16 +150,22 @@
                     <group>
                         <field name="x_product_id" force_save="1" readonly="not x_is_template" invisible="x_targeted_so_id" context="{'default_type': 'service'}"/>
                         <field name="x_reference" readonly="id and not x_is_template"/>
-                        <label for="x_unit_id"/>
-                        <div class="d-flex" name="unit_quantity">
-                            <field name="x_unit_id" class="oe_inline" string="Unit" force_save="1" options="{'no_open': True}" placeholder="m²" invisible="not x_product_id"/>
-                            <field name="x_unit_custom_id" class="oe_inline" string="Unit" force_save="1" options="{'no_open': True}" placeholder="m²" invisible="x_product_id" required="not x_product_id"/>
-                        </div>
+                        <field name="x_unit_id" string="Unit" force_save="1" options="{'no_open': True}" placeholder="m²" invisible="not x_product_id"/>
+                        <field name="x_unit_custom_id" string="Unit" force_save="1" options="{'no_open': True}" placeholder="m²" invisible="x_product_id"/>
```
vava-odoo: missing required from a recent bug fix

construction_developer/data/ir_ui_view.xml
```diff
@@ -150,16 +150,22 @@
                     <group>
                         <field name="x_product_id" force_save="1" readonly="not x_is_template" invisible="x_targeted_so_id" context="{'default_type': 'service'}"/>
                         <field name="x_reference" readonly="id and not x_is_template"/>
-                        <label for="x_unit_id"/>
-                        <div class="d-flex" name="unit_quantity">
-                            <field name="x_unit_id" class="oe_inline" string="Unit" force_save="1" options="{'no_open': True}" placeholder="m²" invisible="not x_product_id"/>
-                            <field name="x_unit_custom_id" class="oe_inline" string="Unit" force_save="1" options="{'no_open': True}" placeholder="m²" invisible="x_product_id" required="not x_product_id"/>
-                        </div>
+                        <field name="x_unit_id" string="Unit" force_save="1" options="{'no_open': True}" placeholder="m²" invisible="not x_product_id"/>
```
vava-odoo:
```suggestion
                        <field name="x_unit_id" force_save="1" options="{'no_open': True}" placeholder="m²" invisible="not x_product_id"/>
```

construction_developer/data/ir_ui_view.xml
```diff
@@ -220,6 +226,61 @@
...
     </record>
+
+    <record id="product_template_form_view_extension" model="ir.ui.view">
+        <field name="arch" type="xml">
+            <button name="action_view_sales" position="after">
+                <button
+                    class="oe_stat_button"
+                    name="%(action_window_see_work_items_from_product_template)d"
+                    type="action" icon="fa-magic"
+                    invisible="x_x_product_template_x_work_items_count &lt;= 0"
```
vava-odoo: I hope it can't be negative
```suggestion
                    invisible="x_x_product_template_x_work_items_count == 0"
```

construction_developer/data/ir_model_fields.xml
```diff
@@ -378,6 +385,30 @@ for record in self:
...
+        <field name="compute"><![CDATA[
+for record in self:
+    record['x_x_product_template_x_work_items_count'] = record.env['x_work_item'].search_count([('x_is_template', '=', True), ('x_work_item_line_ids.x_product_id.product_tmpl_id', '=', record.id)])
+]]>    </field>
+        <field name="ttype">integer</field>
+        <field name="field_description">Work items count</field>
+        <field name="model_id" ref="product.model_product_template"/>
+        <field name="name">x_x_product_template_x_work_items_count</field>
+        <field name="selectable" eval="False"/>
+        <field name="store" eval="False"/>
```
vava-odoo, line 396: and readonly=True

construction_developer/data/ir_model_fields.xml
```diff
@@ -320,9 +320,11 @@ for record in self: record['x_margin'] = (record.x_unit_price - record.x_unit_co
         <field name="compute"><![CDATA[
 for record in self:
     record['x_unit_cost'] = sum(line.x_unit_cost * line.x_quantity for line in record.x_work_item_line_ids)
+    if record.x_is_margin_fixed:
+        record['x_unit_price'] = record.x_unit_cost * (1 + record.x_unit_margin)
 ]]></field>
         <field name="ttype">monetary</field>
-        <field name="depends">x_work_item_line_ids</field>
+        <field name="depends">x_work_item_line_ids.x_unit_cost,x_work_item_line_ids.x_quantity</field>
```
vava-odoo, line 325: why is this change required?

construction_developer/data/ir_model_fields.xml
```diff
@@ -320,9 +320,11 @@ for record in self: record['x_margin'] = (record.x_unit_price - record.x_unit_co
         <field name="compute"><![CDATA[
 for record in self:
     record['x_unit_cost'] = sum(line.x_unit_cost * line.x_quantity for line in record.x_work_item_line_ids)
+    if record.x_is_margin_fixed:
+        record['x_unit_price'] = record.x_unit_cost * (1 + record.x_unit_margin)
```
vava-odoo: From what I recall of my ORM training, it is not a good idea to edit a field in the compute method of another field.
Why not in a server action, like the others?

tests/test_construction_developer/tests/test_action_server.py
```diff
@@ -0,0 +1,88 @@
...
+        self.assertEqual(self.test_template_work_item.x_unit_price, 300.0, "The price of work item is not recomputed correctly when the price of a component changes")
+        self.assertEqual(self.test_template_work_item.x_unit_margin, 2.0, "The margin of work item is not correct after the price of a component changes: it should be recomputed to represent the new value")
+        self.test_template_work_item.x_is_margin_fixed = True
+        self.test_template_work_item.x_unit_price = 200
+        self.assertEqual(self.test_template_work_item.x_unit_margin, 1.0, "When margin is fixed, the margin should be recomputed if we change manually the price of work item")
+        self.test_template_work_item.x_unit_margin = 0.5
+        self.assertEqual(self.test_template_work_item.x_unit_price, 150.0, "When margin is fixed, the price should be recomputed if we change manually the margin of work item")
+        test_template_work_item_lines[1].x_unit_cost = 10
+        self.assertEqual(self.test_template_work_item.x_unit_margin, 0.5, "When margin is fixed, the margin of work item should be stay the same when the cost of a component changes")
```
vava-odoo:
```suggestion
        self.assertEqual(self.test_template_work_item.x_unit_margin, 0.5, "When margin is fixed, the margin of work item should stay the same when the cost of a component changes")
```

vava-odoo (commented):
Good work 💪  and thanks for the tests 🚀 
Mostly small comments

construction_developer/data/base_automation.xml
frva-odoo: You mean instead of on_create_or_write ? It is because it technically does nothing if it is when we create the product:
Case 1; the product is created in a work item line: price and cost are defined when created in the line.
Case 2; the product is created elsewhere: there should be no related work item line existing.
But I could still change it as it isn't doing anything on create it is not bad either

construction_developer/data/ir_model_fields.xml
frva-odoo: I have yet to fully understand how the depends works in this case, but when I tested the code it seemed like changing the quantity on a line without creating/removing one didn't recompute the field. I thought any change in the records would do but it seems like it is computed only when the field changes literally: from x_work_item_line(1, 2) to x_work_item_line(1, 2, 3) for example

construction_developer/data/base_automation.xml
vava-odoo: That is from the UI, but what about another server action creating it?
Eventhough it's unlikely, I would safeguard it, yes.

tests/test_construction_developer/__manifest__.py
```diff
@@ -0,0 +1,9 @@
+{
+    'name': 'Test Construction Developer',
+    'version': '1.0',
```
vava-odoo, line 3: manifest bump in construction_developer disappeared with the rebase. I definitely need to add a test on this (task is ongoing)

construction_developer/data/ir_actions_server.xml
```diff
@@ -135,17 +145,45 @@ record.write({'x_unit_price': record.x_product_id.list_price, 'x_unit_cost': rec
...
-    'tag': 'display_notification',
-    'params': {'message': record.x_name + " template saved.", 'type': 'success'},
-}
+for wi in records:
+    wi.copy(default={'x_sale_order_line_id': False, 'x_is_template': True, 'x_targeted_so_id': False})
+    action = {
+        'type': 'ir.actions.client',
+        'tag': 'display_notification',
+        'params': {'message': wi.x_name + " template saved.", 'type': 'success'},
+    }
```
vava-odoo: question: does it display for all modified records 🤔 
maybe something like this?
```suggestion
action = {
    'type': 'ir.actions.client',
    'tag': 'display_notification',
    'params': {'message': ', '.join(wi.x_name for wi in records) + " template saved.", 'type': 'success'},
}
```

construction_developer/data/ir_model_fields.xml
vava-odoo: well spotted!

vava-odoo (approved):
Well done 💪

PR #1912, opened 2026-03-27
[IMP] booking_engine: add availabilities

trcazier:
Adds a new feature for the booking engine: availability & occupancy, which allow the user to monitor the occupancy of their planning resources for each day using pills in a color-coded Gantt chart heatmap.

This is done with automations on create & update of several models tied to the computation of an availability, namely resource.calendar.leaves, planning.slot, product.template & resource.resource. A CRON is also scheduled at night to create the availability records for the next 500 days in prospect of the channex integration.

In more detail, a single update server action is used across all the automations' server actions to recompute all the availabilities flagged with a `x_to_recompute` boolean to minimize faulty states due to race conditions. This way, each automation only has to flag the current (and new availabilites in the case of on create or edit) availabilities tied to records being updated or unlinked with that boolean, then trigger the recomputation in a separate server action that will fetch all the availabilities that have to be recomputed instead of each automation recomputing it separately with its own state.
A CRON is used for the unlinks instead of just a server action because calling the server action directly does not have the desired behavior since the server action is triggered before the records are actually deleted. Therefore, the records to be deleted are still taken into account for the recomputation. Using a CRON fixes this because the computation happens after the records are removed from the DB. Unlinks of these 4 models are rare cases, and therefore less likely to cause a problem than running the CRON for every little change, hence why we still use the server action for those, in addition to it working properly unlike the unlink.

Also adds more demo data for each of the industries using booking_engine.

task-5913061

### DIFF
--- a/booking_engine/data/base_automation.xml
+++ b/booking_engine/data/base_automation.xml
@@ -80,4 +80,83 @@
         <field name="filter_pre_domain">[('project_id.x_is_house_keeping_project', '!=', False)]</field>
         <field name="trigger_field_ids" eval="[(6, 0, [ref('project.field_project_task__stage_id')])]"/>
     </record>
+
+    <!--
+        Recomputation of availabilities
+    -->
+    <!-- planning.slot -->
+    <record id="automation_on_planning_slot_change" model="base.automation">
+        <field name="name">Booking Engine: Mark Availabilities To Recompute: Slot Change</field>
+        <field name="model_id" ref="planning.model_planning_slot"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [
+            ref('planning.field_planning_slot__start_datetime'), 
+            ref('planning.field_planning_slot__end_datetime'), 
+            ref('planning.field_planning_slot__resource_id')
+        ])]"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('booking_engine.server_action_update_availability_for_planning_slot')])]"/>
+    </record>
+    <record id="automation_on_planning_slot_delete" model="base.automation">
+        <field name="name">Booking Engine: Mark Availabilities To Recompute: Slot Delete</field>
+        <field name="model_id" ref="planning.model_planning_slot"/>
+        <field name="trigger">on_unlink</field>
+        <field name="action_server_ids" eval="[(6, 0, [ref('booking_engine.server_action_update_availability_for_planning_slot_delete')])]"/>
+    </record>
+
+    <!-- resource.calendar.leave -->
+    <record id="automation_on_resource_calendar_leaves_change" model="base.automation">
+        <field name="name">Booking Engine: Mark Availabilities To Recompute: Leave Change</field>
+        <field name="model_id" ref="resource.model_resource_calendar_leaves"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [
+            ref('resource.field_resource_calendar_leaves__date_from'), 
+            ref('resource.field_resource_calendar_leaves__date_to'), 
+            ref('resource.field_resource_calendar_leaves__resource_id'), 
+            ref('resource.field_resource_calendar_leaves__calendar_id')
+        ])]"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('booking_engine.server_action_update_availability_for_resource_calendar_leaves')])]"/>
+    </record>
+    <record id="automation_on_resource_calendar_leaves_delete" model="base.automation">
+        <field name="name">Booking Engine: Mark Availabilities To Recompute: Leave Delete</field>
+        <field name="model_id" ref="resource.model_resource_calendar_leaves"/>
+        <field name="trigger">on_unlink</field>
+        <field name="action_server_ids" eval="[(6, 0, [ref('booking_engine.server_action_update_availability_for_resource_calendar_leaves_delete')])]"/>
+    </record>
+
+    <!-- resource.resource -->
+    <record id="automation_on_resource_resource_change" model="base.automation">
+        <field name="name">Booking Engine: Mark Availabilities To Recompute: Room Change</field>
+        <field name="model_id" ref="resource.model_resource_resource"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [
+            ref('resource.field_resource_resource__active'), 
+            ref('resource.field_resource_resource__calendar_id')
+        ])]"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('booking_engine.server_action_update_availability_for_resource_resource')])]"/>
+    </record>
+    <record id="automation_on_resource_resource_delete" model="base.automation">
+        <field name="name">Booking Engine: Mark Availabilities To Recompute: Room Delete</field>
+        <field name="model_id" ref="resource.model_resource_resource"/>
+        <field name="trigger">on_unlink</field>
+        <field name="action_server_ids" eval="[(6, 0, [ref('booking_engine.server_action_update_availability_for_resource_resource_delete')])]"/>
+    </record>
+
+    <!-- product.template -->
+    <record id="automation_on_product_template_change" model="base.automation">
+        <field name="name">Booking Engine: Mark Availabilities To Recompute: Offer Change</field>
+        <field name="model_id" ref="product.model_product_template"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [
+            ref('booking_engine.product_model_product_template_x_resource_ids_field'), 
+            ref('booking_engine.product_model_product_template_x_is_a_room_offer_field'), 
+            ref('product.field_product_template__active')
+        ])]"/>
+        <field name="action_server_ids" eval="[(6, 0, ref('booking_engine.server_action_update_availability_for_product_template'))]"/>
+    </record>
+    <record id="automation_on_product_template_delete" model="base.automation">
+        <field name="name">Booking Engine: Mark Availabilities To Recompute: Offer Delete</field>
+        <field name="model_id" ref="product.model_product_template"/>
+        <field name="trigger">on_unlink</field>
+        <field name="action_server_ids" eval="[(6, 0, [ref('booking_engine.server_action_update_availability_for_product_template_delete')])]"/>
+    </record>
 </odoo>
--- a/booking_engine/data/ir_actions_server.xml
+++ b/booking_engine/data/ir_actions_server.xml
@@ -5,8 +5,7 @@
         <field name="model_id" ref="x_model_city_tax"/>
         <field name="state">code</field>
         <field name="code"><![CDATA[
-stay_tax_line = False
-sequence = 0
+sequence, stay_tax_line = 0, False
 for so_line in record.x_sale_order_id.order_line:
     if so_line.product_id.x_is_stay_tax: stay_tax_line = so_line
     sequence = max(sequence, so_line.sequence)
@@ -14,7 +13,7 @@ if stay_tax_line: stay_tax_line.write({'product_uom_qty': record.x_total, 'qty_d
 elif record.x_total > 0:
     product = env['product.product'].search([('x_is_stay_tax', '=', True)], limit=1)
     env['sale.order.line'].create({'order_id': record.x_sale_order_id.id, 'product_id': product.id, 'product_uom_qty': record.x_total, 'qty_delivered': record.x_total})
-        ]]></field>
+]]></field>
     </record>
     <record id="open_x_city_tax_action" model="ir.actions.act_window">
         <field name="name">City Tax</field>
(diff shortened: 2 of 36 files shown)


booking_engine/data/ir_actions_server.xml
```diff
@@ -197,6 +196,257 @@ if (project := env['project.project'].search([('x_is_house_keeping_project', '='
...
+
+    <!-- Availability dependencies: Delete-->
+    <record id="server_action_update_availability_for_planning_slot_delete" model="ir.actions.server">
+        <field name="name">Booking Engine: Slot Delete Availability Logic</field>
+        <field name="model_id" ref="planning.model_planning_slot"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+if (to_recompute := records.mapped('x_availability_ids')): to_recompute.write({'x_to_recompute': True})
+
+env.ref('booking_engine.ir_cron_update_availabilities')._trigger()
```
vava-odoo: Shouldn't this be triggered only if `to_recompute` is not empty?

booking_engine/data/ir_actions_server.xml
```diff
@@ -197,6 +196,257 @@ if (project := env['project.project'].search([('x_is_house_keeping_project', '='
...
+if (offers := env['product.template'].search([('x_is_a_room_offer', '=', True)])):
+    today, target_dates = datetime.date.today(), set(datetime.date.today() + datetime.timedelta(days=i) for i in range(500))
+
+    # Fetch all existing records in that window
+    existing_avails = env['x_availability'].search([('x_stay_offer', 'in', offers.ids), 
+        ('x_date', '>=', today), ('x_date', '<', today + datetime.timedelta(days=500))])
+
+    to_create = []
+    for offer in offers:
+        for d in (target_dates - set(existing_avails.filtered(lambda a: a.x_stay_offer.id == offer.id).mapped('x_date'))):
```
vava-odoo: This loop in a loop with both filtered and mapped is a performance killer. Maybe you could gain performance by using _read_group instead of search, such that you group all existing dates by offer in one call.

booking_engine/data/ir_actions_server.xml
```diff
@@ -197,6 +196,257 @@ if (project := env['project.project'].search([('x_is_house_keeping_project', '='
...
+            to_create.append({'x_stay_offer': offer.id, 'x_date': d, 'x_to_recompute': True})
+
+    if to_create: env['x_availability'].create(to_create)
+    env['x_availability'].search([('x_date', '>=', today)]).write({'x_to_recompute': True})
```
vava-odoo, line 304: do you need to recompute if none was created? 🤔

booking_engine/data/ir_model_access.xml
```diff
@@ -18,4 +18,13 @@
...
         <field name="perm_unlink" eval="True"/>
     </record>
+    <record id="x_model_availability_access" model="ir.model.access">
+        <field name="group_id" ref="base.group_user"/>
+        <field name="model_id" ref="x_model_availability"/>
+        <field name="name">x_model_availability_group_user_access</field>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
```
vava-odoo: don't we want to restrict potential mess here? Only reading is needed for a user, all other operations should be done in sudo IMO.

booking_engine/data/ir_model_fields.xml
```diff
@@ -636,4 +634,150 @@ for record in self:
         <field name="store" eval="False"/>
         <field name="related">x_resource_id.x_occupancy_color</field>
     </record>
+  <record id="x_name_field_x_availability" model="ir.model.fields">
+    <field name="ttype">char</field>
+    <field name="copied" eval="True"/>
+    <field name="field_description">Description</field>
+    <field name="model_id" ref="x_model_availability"/>
+    <field name="name">x_name</field>
+    <field name="on_delete" eval="False"/>
```
vava-odoo: on_delete only makes sense with linked records

booking_engine/data/ir_model_fields.xml
```diff
@@ -636,4 +634,150 @@ for record in self:
...
+    <field name="translate">standard</field>
+  </record>
+  <record id="x_active_field_x_availability" model="ir.model.fields">
+    <field name="ttype">boolean</field>
+    <field name="copied" eval="True"/>
+    <field name="field_description">Active</field>
+    <field name="model_id" ref="x_model_availability"/>
+    <field name="name">x_active</field>
+    <field name="on_delete" eval="False"/>
+  </record>
```
vava-odoo, line 649: I wonder if x_name and x_active are not there by default, but not used at all (except in default views)

booking_engine/data/ir_model_fields.xml
```diff
@@ -636,4 +634,150 @@ for record in self:
...
+  <record id="x_booked_field_x_availability" model="ir.model.fields">
+    <field name="ttype">integer</field>
+    <field name="copied" eval="True"/>
+    <field name="field_description">Booked</field>
+    <field name="model_id" ref="x_model_availability"/>
+    <field name="name">x_booked</field>
+    <field name="on_delete" eval="False"/>
+  </record>
+  <record id="x_available_field_x_availability" model="ir.model.fields">
+    <field name="compute"><![CDATA[for record in self: record['x_available'] = record.x_total_units - record.x_booked]]></field>
```
vava-odoo, line 694: does this one need to be stored? I don't think but I may have missed something

booking_engine/data/ir_default.xml
```diff
@@ -0,0 +1,31 @@
...
+    <field name="json_value">"0"</field>
+  </record>
+  <record id="default_x_available_field_x_availability" model="ir.default">
+    <field name="field_id" ref="x_available_field_x_availability"/>
+    <field name="json_value">"0"</field>
+  </record>
+  <record id="default_x_occupancy_color_field_x_availability" model="ir.default">
+    <field name="field_id" ref="x_occupancy_color_field_x_availability"/>
+    <field name="json_value">"0"</field>
+  </record>
```
vava-odoo: Do you need a default value for non-stored fields?

booking_engine/data/ir_model_fields.xml
```diff
@@ -636,4 +634,150 @@ for record in self:
...
+  <record id="x_occupancy_field_x_availability" model="ir.model.fields">
+    <field name="compute"><![CDATA[for record in self:
+      record['x_occupancy'] = 0 if not record.x_total_units else 1 - record.x_available / record.x_total_units ]]></field>
+    <field name="ttype">float</field>
+    <field name="depends">x_available,x_total_units</field>
+    <field name="field_description">Occupancy</field>
+    <field name="model_id" ref="x_model_availability"/>
+    <field name="name">x_occupancy</field>
+    <field name="on_delete" eval="False"/>
+    <field name="readonly" eval="True"/>
```
vava-odoo, line 723: why only this one is readonly. From what I understand, none of these fields should be editable from the interface (only done by the actions)

booking_engine/data/ir_ui_view.xml
```diff
@@ -823,4 +823,206 @@
...
+    <field name="type">form</field>
+    <field name="active" eval="True"/>
+    <field name="arch" type="xml">
+      <form>
+        <header/>
+        <sheet string="Availability">
+          <widget name="web_ribbon" text="Archived" bg_color="text-bg-danger" invisible="x_active"/>
+          <field name="x_active" invisible="1"/>
+          <group name="studio_group_390012">
+            <group name="studio_group_390012_left">
```
vava-odoo:
```suggestion
          <group>
            <group>
```

booking_engine/data/ir_ui_view.xml
```diff
@@ -823,4 +823,206 @@
...
+              <field name="x_date"/>
+              <field name="x_stay_offer" domain="[(&quot;x_is_a_room_offer&quot;, &quot;!=&quot;, False)]"/>
```
vava-odoo: Already defined on the field itself
```suggestion
              <field name="x_stay_offer"/>
```

booking_engine/data/ir_ui_view.xml
```diff
@@ -823,4 +823,206 @@
...
+              <field name="x_total_units"/>
+              <field name="x_booked"/>
+              <field name="x_available"/>
+              <field name="x_occupancy" widget="percentage"/>
+            </group>
+            <group name="studio_group_390012_right"/>
```
vava-odoo:
```suggestion
```

booking_engine/data/ir_ui_view.xml
```diff
@@ -823,4 +823,206 @@
...
+    <field name="active" eval="True"/>
+    <field name="arch" type="xml">
+      <search>
+        <field name="x_name"/>
+        <separator/>
+        <filter string="Archived" name="archived_x_availability" domain="[['x_active', '=', False]]"/>
+        <separator/>
+      </search>
+    </field>
+  </record>
```
vava-odoo, line 885: this looks like an automatically generated one. Maybe delete unnecessary ones

booking_engine/data/ir_model_fields.xml
```diff
@@ -636,4 +634,150 @@ for record in self:
...
+    <field name="selectable" eval="False"/>
+    <field name="store" eval="False"/>
+  </record>
+  <record id="x_occupancy_color_field_x_availability" model="ir.model.fields">
+    <field name="ttype">integer</field>
+    <field name="depends">x_occupancy,x_total_units</field>
```
vava-odoo: a bit weird to depend on x_occupancy and see x_availability in the formula

booking_engine/data/ir_ui_view.xml
```diff
@@ -823,4 +823,206 @@
...
+    ]"/>
+  </record>
+
+  <record id="view_x_availability_form_inherit" model="ir.ui.view">
+    <field name="model">x_availability</field>
+    <field name="name">x_availability.view.form.inherit</field>
+    <field name="type">form</field>
+    <field name="inherit_id" ref="view_x_availability_form"/>
+    <field name="priority">99</field>
+    <field name="mode">extension</field>
```
vava-odoo: don't extend in the same file, this is just the history of the studio modification. Just move this on the first definition above

booking_engine/data/ir_ui_view.xml
```diff
@@ -823,4 +823,206 @@
...
+          <field name="x_availability_color"/>
+        </xpath>
+        <xpath expr="/form//field[@name='x_occupancy']" position="after">
+          <field name="x_occupancy_color"/>
+        </xpath>
+      </data>
+    </field>
+  </record>
+
+  <record id="view_x_availability_kanban" model="ir.ui.view">
```
vava-odoo: Where is this kanban displayed?

booking_engine/data/ir_ui_view.xml
```diff
@@ -823,4 +823,206 @@
...
+        <field name="x_booked"/>
+        <field name="x_date"/>
+        <field name="x_total_units"/>
+        <templates>
+          <div t-name="gantt-popover">
+            <t t-set="bar_color" t-value="x_availability_color"/>
+            <t t-set="bar_pct" t-value="x_availability"/>
+            <t t-set="main_title" t-value="'Availability:'"/>
+            
+            <t t-set="heat_scale" t-value="['#ffdc00', '#fbc904', '#f6b609', '#f2a30d', '#ed9011', '#e97d16', '#e4691a', '#e0561e', '#db4323', '#d73027']"/>
```
vava-odoo, line 909: why do you need both this heat scale and the scss definition?

campsite/data/views.xml
vava-odoo: From what I read, this is useless as the field description is already that one in booking_engine
```suggestion
```

guest_house/data/knowledge_article.xml
```diff
@@ -289,6 +289,25 @@
                 class="btn btn-secondary">🎓 Invoicing</a>
         </p>
         <p><br/></p>
+        <h3>🚀 Strategic Oversight</h3>
+        <hr/>guest house
```
vava-odoo:
```suggestion
        <hr/>
```

booking_engine/data/ir_actions_server.xml
```diff
@@ -197,6 +196,257 @@ if (project := env['project.project'].search([('x_is_house_keeping_project', '='
...
+
+    <!-- Availability dependencies: Update -->
+    <record id="server_action_update_availability_for_planning_slot" model="ir.actions.server">
+        <field name="name">Booking Engine: Availability Slot Modify Logic</field>
+        <field name="model_id" ref="planning.model_planning_slot"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+# Mark all currently linked availabilities as to recompute
+# This catches the old dates AND the old resource if the user dragged it to a different resource
+if (old := records.mapped('x_availability_ids')): old.write({'x_to_recompute': True})
```
vava-odoo: why mapped? Doesn't records.x_availability_ids work as well?

booking_engine/data/ir_actions_server.xml
```diff
@@ -197,6 +196,257 @@ if (project := env['project.project'].search([('x_is_house_keeping_project', '='
...
+        <field name="model_id" ref="resource.model_resource_calendar_leaves"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+# Mark all currently linked availabilities as to recompute
+# This catches the old dates AND the old resources if the user changed them
+if (old := records.mapped('x_availability_ids')): old.write({'x_to_recompute': True})
+
+undirect = records - (direct := records.filtered('resource_id'))
+
+if (resources := env['resource.resource'].search(['|', ('calendar_id', 'in', undirect.mapped('calendar_id').ids), ('id', 'in', direct.mapped('resource_id').ids)])):
```
vava-odoo: same, why mapped?

booking_engine/data/ir_actions_server.xml
```diff
@@ -197,6 +196,257 @@ if (project := env['project.project'].search([('x_is_house_keeping_project', '='
...
+    </record>
+
+    <record id="server_action_update_availability_for_resource_resource_delete" model="ir.actions.server">
+        <field name="name">Booking Engine: Room Delete Availability Logic</field>
+        <field name="model_id" ref="resource.model_resource_resource"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+# Rooms don't have direct many2many links to availabilities but they link through the stay offer
+if (offers := env['product.template'].search([('x_resource_ids', 'in', records.ids)])):
+    if (avails := env['x_availability'].search([('x_stay_offer', 'in', offers.ids)])): avails.write({'x_to_recompute': True})
```
vava-odoo: Why not this instead?
```suggestion
if (avails := env['x_availability'].search([('x_stay_offer.x_resource_ids', 'in', records.ids)])): avails.write({'x_to_recompute': True})
```
And actually, if you don't care about `avails`, you could simply write this (write just silently voids if the recordset is empty)
```suggestion
env['x_availability'].search([('x_stay_offer.x_resource_ids', 'in', records.ids)]).write({'x_to_recompute': True})
```

booking_engine/data/ir_actions_server.xml
```diff
@@ -197,6 +196,257 @@ if (project := env['project.project'].search([('x_is_house_keeping_project', '='
...
+    </record>
+
+    <record id="server_action_update_availability_for_resource_resource" model="ir.actions.server">
+        <field name="name">Booking Engine: Availability Room Modify Logic</field>
+        <field name="model_id" ref="resource.model_resource_resource"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+if offers := env['product.template'].search([('x_resource_ids', 'in', records.ids)]):
+    # Find all FUTURE availabilities
+    if avails := env['x_availability'].search([('x_stay_offer', 'in', offers.ids), ('x_date', '>=', datetime.date.today())]):
```
vava-odoo: same as above

vava-odoo: also: wouldn't it be interesting to archive past availabilities?

booking_engine/data/base_automation.xml
```diff
@@ -80,4 +80,83 @@
...
+    <record id="automation_on_product_template_change" model="base.automation">
+        <field name="name">Booking Engine: Mark Availabilities To Recompute: Offer Change</field>
+        <field name="model_id" ref="product.model_product_template"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [
+            ref('booking_engine.product_model_product_template_x_resource_ids_field'), 
+            ref('booking_engine.product_model_product_template_x_is_a_room_offer_field'), 
+            ref('product.field_product_template__active')
+        ])]"/>
+        <field name="action_server_ids" eval="[(4, ref('booking_engine.server_action_update_availability_for_product_template'))]"/>
```
vava-odoo: why 4 (link) and not 6,0 (set) like others?

booking_engine/data/ir_actions_server.xml
```diff
@@ -197,6 +196,257 @@ if (project := env['project.project'].search([('x_is_house_keeping_project', '='
...
+        env.ref('booking_engine.server_action_update_availabilities').run()
+]]></field>
+    </record>
+
+    <record id="server_action_update_availability_for_product_template" model="ir.actions.server">
+        <field name="name">Booking Engine: Availability Offer Modify Logic</field>
+        <field name="model_id" ref="product.model_product_template"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+if (offers := records.filtered('x_is_a_room_offer')):
```
vava-odoo: the filtered should be a domain on the automation

vava-odoo (commented):
Huge work! We'll probably need to discuss some parts, let me know.

booking_engine/data/ir_actions_server.xml
trcazier: This is the cron that is triggered every day at night that creates new ones but also recomputes all the availabilities to correct the occasional race conditions/erroneous states, so yes

booking_engine/data/ir_model_fields.xml
trcazier: So you mean I can remove them?

booking_engine/data/ir_ui_view.xml
trcazier: It was used before to modify the popover that was displayed in the gantt (yes, the gantt, from something in the kanban view...) but I managed to find how to do it in the gantt and forgot to remove it

booking_engine/data/ir_ui_view.xml
trcazier: The SCSS is to override the gantt pills' default color. It dynamically creates the CSS classes `o_gantt_color_{n}` based on the field you provide as the `color=""` attribute in `<gantt>` and assigns it a color based on the `int` in the color field, meaning you can't pick any color you want yourself. The heat scale in the view is there to change the color in the progress bar in the popover.

booking_engine/data/ir_actions_server.xml
trcazier: Why does that work? When does it work and when does it not for multiple records? I usually do mapped because I don't trust accessing records with . when there are multiple ones because sometimes it doesn't work. For example, if I replace every `.mapped('field')` in this file with .field I get "expected singleton"

booking_engine/data/ir_actions_server.xml
```diff
@@ -273,32 +275,82 @@ if (records := env['x_availability'].search([('x_to_recompute', '=', True)])):
         <field name="state">code</field>
         <field name="usage">ir_cron</field>
         <field name="code"><![CDATA[
+# Archive all past availabilities (not important to see)
+env['x_availability'].search([('x_date', '<', today), ('x_active', '=', True)]).write({'x_active': False})
```
vava-odoo: just validated with Pierre, well done. Almost 😄 
```suggestion
env['x_availability'].search([('x_date', '<', 'today'), ('x_active', '=', True)]).write({'x_active': False})
```

booking_engine/data/ir_actions_server.xml
```diff
@@ -273,32 +275,82 @@ if (records := env['x_availability'].search([('x_to_recompute', '=', True)])):
...
-        ('x_date', '>=', today), ('x_date', '<', today + datetime.timedelta(days=500))])
+    existing_dates_by_offer = {offer_id: set() for offer_id in offers.ids}
+
+    # bypassing heavy ORM recordsets
+    raw_avails = env['x_availability'].search_read(
+        [('x_stay_offer', 'in', offers.ids), ('x_date', '>=', today), ('x_date', '<', today + datetime.timedelta(days=500))],
+        ['x_stay_offer', 'x_date'])
+
+    for data in raw_avails:
+        existing_dates_by_offer[data['x_stay_offer'][0]].add(data['x_date'])
```
vava-odoo: Probably better with a _read_group
```suggestion
    existing_dates_by_offer = env['x_availability']._read_group(
        [('x_stay_offer', 'in', offers.ids), ('x_date', '>=', today), ('x_date', '<', today + datetime.timedelta(days=500))],
        groupby=['x_stay_offer'], aggregates=['x_date:array_agg']
    )
```

booking_engine/data/ir_actions_server.xml
```diff
@@ -418,13 +430,11 @@ if (resources := env['resource.resource'].search(['|', ('calendar_id', 'in', und
...
         <field name="state">code</field>
         <field name="code"><![CDATA[
-if offers := env['product.template'].search([('x_resource_ids', 'in', records.ids)]):
-    # Find all FUTURE availabilities
-    if avails := env['x_availability'].search([('x_stay_offer', 'in', offers.ids), ('x_date', '>=', datetime.date.today())]):
-        # Mark them to recompute
-        avails.write({'x_to_recompute': True})
+# Mark all currently linked availabilities as to recompute
+env['x_availability'].search([('x_stay_offer.x_resource_ids', 'in', records.ids), ('x_date', '>=', datetime.date.today())]) \
+    .write({'x_to_recompute': True})
```
vava-odoo: we need to discuss the line break convention

booking_engine/data/ir_actions_server.xml
```diff
@@ -273,32 +275,82 @@ if (records := env['x_availability'].search([('x_to_recompute', '=', True)])):
...
+
+# Compute target offers & link
+for rec in records.filtered(lambda r: r.id in dates):
+    if records._name == 'planning.slot':
+        target_res_ids = [rec.resource_id.id]
+    else: # resource.calendar.leaves
+        target_res_ids = [rec.resource_id.id] if rec.resource_id else resources.filtered(lambda r: r.calendar_id.id == rec.calendar_id.id).ids
+
+    if (target_offers := offers.filtered(lambda o: set(target_res_ids) & set(o.x_resource_ids.ids))):
+        new_avails = avails.filtered(lambda a: a.x_stay_offer.id in target_offers.ids and a.x_date in dates[rec.id])
```
vava-odoo: I don't get the filter on dates. You need to link all availabilities from this stay offer, no?

booking_engine/data/ir_actions_server.xml
vava-odoo: I see that I already suggested _read_group here. Was there something I missed that prevents using it?

booking_engine/data/ir_model_access.xml
vava-odoo: I'm stuck here. You removed the write and create access rights on user, but did not add any sudo nor system user access right. What is the trick?

booking_engine/data/ir_ui_view.xml
```diff
@@ -852,30 +853,32 @@
...
               <field name="x_total_units"/>
               <field name="x_booked"/>
               <field name="x_available"/>
+              <field name="x_availability" widget="percentage"/>
+              <field name="x_availability_color"/>
               <field name="x_occupancy" widget="percentage"/>
+              <field name="x_occupancy_color"/>
             </group>
-            <group name="studio_group_390012_right"/>
+            <group/>
```
vava-odoo: do we need a second group? 🤔

booking_engine/data/ir_model_fields.xml
vava-odoo: x_active is now used, so no. x_name, I think you can remove it.

booking_engine/data/ir_actions_server.xml
trcazier: The way this code works is the following:
1. To save ORM calls, we first compute all the dates related to the records passed in the arguments (which are either leaves or slots), because we only want to fetch the availabilities once. We'll then have to "assign" the availabilities later.

```python
dates, unique_dates = {}, set()
for rec in records:
    if not rec[start_f] or not rec[end_f]: continue
    d_min, d_max = rec[start_f].date(), rec[end_f].date()
    dates[rec.id] = [d_min + datetime.timedelta(days=i) for i in range((d_max - d_min).days + 1)]
    unique_dates.update(dates[rec.id])

```

2. Then, we fetch the resources related to our arguments.

```python
offers = env['product.template'].search([('x_resource_ids', 'in', resources.ids)]) if resources else []
```

3. Then, we fetch all the potential availabilities related to the dates we computed and the resources we have.
```python
avails = env['x_availability'].search([('x_date', 'in', list(unique_dates)), ('x_stay_offer', 'in', offers.ids)]) if unique_dates and offers else env['x_availability']
```

4. Now, for all the records for which we have dates, we're gonna have to find back all the information we had to compress into one query
```python
for rec in records.filtered(lambda r: r.id in dates):
```

5. We get the resources corresponding to this specific record
```python
    if records._name == 'planning.slot':
        target_res_ids = [rec.resource_id.id]
    else: # resource.calendar.leaves
        target_res_ids = [rec.resource_id.id] if rec.resource_id else resources.filtered(lambda r: r.calendar_id.id == rec.calendar_id.id).ids
```

6. And we get the availabilities corresponding to this record, resource and date
```python
    if (target_offers := offers.filtered(lambda o: set(target_res_ids) & set(o.x_resource_ids.ids))):
        new_avails = avails.filtered(lambda a: a.x_stay_offer.id in target_offers.ids and a.x_date in dates[rec.id])
```

7. And we assign them to this record
```python
        rec.write({'x_availability_ids': [(6, 0, new_avails.ids)]})
        new_avails_to_recompute |= new_avails
```

Now picture this:
Let's say in our records we have 
- rec1: one slot on resource A from 2026-01-01 to 2026-01-31
- rec2: one slot on resource A from 2027-01-01 to 2027-01-31

If we didn't do the filter, it would assign all availabilities from BOTH 2026-01 and 2027-01 to both rec1 and rec2 instead of only the ones corresponding to their respective dates.

booking_engine/data/ir_ui_view.xml
trcazier: I mean we're not even supposed to access this view anywhere so why bother

booking_engine/data/ir_model_access.xml
trcazier: Read/write access do not seem to matter for the cron or base automations, so this works just fine, since the user is not supposed to interact directly with the availabilities records in any way. Leaving it like this

booking_engine/data/ir_actions_server.xml
```diff
@@ -197,6 +196,333 @@ if (project := env['project.project'].search([('x_is_house_keeping_project', '='
...
+if (offers := env['product.template'].search([('x_is_a_room_offer', '=', True)])):
+    existing_dates_by_offer = env['x_availability']._read_group(
+            [('x_stay_offer_id', 'in', offers.ids), ('x_date', '>=', today), ('x_date', '<', today + datetime.timedelta(days=500))],
+            groupby=['x_stay_offer_id'], aggregates=['x_date:array_agg']
+        )
+
+    # Create availabilities only for dates that don't exist
+    to_create = [{'x_stay_offer_id': offer_id, 'x_date': d, 'x_to_recompute': True}
+                 for offer_id in offers.ids
+                 for d in (target_dates - existing_dates_by_offer[offer_id])]
```
vava-odoo: _read_group will not fill the dict for that key if there is none (unlike your previous version with default empty set for all)
```suggestion
                 for d in (target_dates - existing_dates_by_offer.get(offer_id, set())])]
```

booking_engine/data/ir_actions_server.xml
```diff
@@ -197,6 +196,333 @@ if (project := env['project.project'].search([('x_is_house_keeping_project', '='
...
+    <record id="server_action_update_availability_for_product_template_delete" model="ir.actions.server">
+        <field name="name">Booking Engine: Stay Offer Delete Availability Logic</field>
+        <field name="model_id" ref="product.model_product_template"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+# If a stay offer is deleted, we can delete the entirety of its availabilities because
+# we can't see that line in the Gantt chart anymore
+env['x_availability'].search([('x_stay_offer_id', 'in', records.ids)]).unlink()
+
+# Use the cron so that the current records (to delete) aren't taken into account for the recomputation
```
vava-odoo: why do we need to run the cron then? Since we deleted the whole row. Other rows are unaffected

vava-odoo (approved):

vava-odoo:
did a small fixup commit. Feel free to squash if ok for you. 
I just wonder if tiny tests like just running every server actions is necessary. Just because some field names changed from 19 to master (resource_id -> resource_ids in 19.2 for example).

vava-odoo:
@robodoo r+

PR #1933, opened 2026-03-30
[FIX] construction_developer: assign variant price to work item line

frva-odoo:
Before this commit, in the server action server_action_set_unit_cost_price_from_product, we used the list_price field of the product to assign to the price of the work item line. 
This field is defined in the product template, which does not take into account extra price from attribute values, nor custom price assigned on the variant directly.
This commit fixes it by assigning the variant price, lst_price


vava-odoo (approved):
@robodoo r+ squash

PR #1934, opened 2026-03-30
[FIX] test_industry: Align CLI behavior by removing active_test=False

amas-odoo:
In the industry CLI, the environment was initialized with `active_test=False`, which disables the default filtering of inactive (archived) records. This caused searches to return archived records as well, leading to incorrect behavior such as creating quality checks from archived quality points in `Food Distribution` module.

Remove `active_test=False` to restore the default behavior where only active records are considered.

Task ID: 6076908


vava-odoo (approved):
@robodoo r+

PR #1939, opened 2026-03-30
[FIX] *: assigning correct model object

vava-odoo:
In several cases, there is a mismatch between compared/assigned record models, leading to erroneous behaviour. This commit ensures assignation and domain criterion is correct at the model level.

Furthermore, for HVAC, the wizard is adapted to use product.product, instead of product.template.

backport of odoo/industry#1844


vava-odoo:
@robodoo r+ fw=no

PR #1956, opened 2026-04-01
[FIX] *: add context key for product variant creation

trcazier:
This commit https://github.com/odoo/odoo/pull/254323 changed the way product( template)s are created, which now decouples the logic into two context attributes instead of one. This commit fixes this by adding the second one.


vava-odoo (approved):
@robodoo r+
FYI @abdrahmanrashed

PR #1972, opened 2026-04-02
[IMP] condominium: remove VAT dependency for condominium button visibility

mega-odoo:
`Before this commit`

The `Create Condominium` button in `res.partner` form view is visible based on the presence of a VAT number to identify a company.

`After this commit`

The button is now displayed only when both `ref_company_ids` and `x_companies` are empty.

`Purpose:`

VAT is a real business field, but here it's being used as a dummy value. instead of using a dummy VAT number, we want to change the logic to decide when to show the button properly.

TASK: 6064101


vava-odoo (approved):
@robodoo r+

PR #1976, opened 2026-04-03
[FIX] booking_engine: fix product form view layout

dhrs-odoo:
Previously, only the label of lst_price was targeted via XPath, which was insufficient and resulted in a broken or inconsistent layout.

This commit enhances the fix by also applying an XPath on list_price_uom, ensuring the UI structure remains consistent and the design is properly aligned.


vava-odoo (approved):
@robodoo r+

PR #1977, opened 2026-04-03
[FIX] booking_engine: fix product form view layout

dhrs-odoo:
Previously, the label of `list_price` was targeted via XPath, which was broken due to standard changes.

This commit updates the fix by targeting `lst_price` instead of `list_price`, and adds an additional XPath on `list_price_per_uom_id` to preserve the layout and ensure a consistent UI.

Related PR - https://github.com/odoo/odoo/pull/257417


vava-odoo (approved):
@robodoo r+

PR #1983, opened 2026-04-09
[FIX] real_estate: store related address fields to prevent cache issues

dhrs-odoo:
Before this commit, the fields x_street, x_city, x_zip_code, and x_country were unstored on properties (products), and values were only stored on the linked partner. When creating a new property, these fields values were cached and added to the technical partner after its creation by an automation.

Due to recent changes in the ORM*, these fields were not cached anymore, leading to blank values.

This commit sets these related fields to stored ones to ensure their values are properly stored.

*https://github.com/odoo/odoo/commit/42df5cf312dcc4947cc9065d36fe1ed52145e016


vava-odoo:
@robodoo override=ci/industry
@robodoo r+
Thanks!

PR #1988, opened 2026-04-13
[CLN] tests: default test tag is post_install

vava-odoo:
Since saas-19.1*, default tag for tests is post_install, where it was at_install before.
This commit removes the useless post_install redefinition of the tag.

*https://github.com/odoo/odoo/commit/18a725879777699984d77211f523171dad28d881


vava-odoo:
@robodoo r+

PR #2020, opened 2026-04-20
[IMP] booking_engine: add resource tag on partner view from pos

vava-odoo:
Before this commit, the new light form view of res.partner accessible from the POS was not extended to display the new custom fields. Since the resource (the room number) is an important information to display, this commit adds it to the view.

task-6126922


vava-odoo:
@robodoo r+

vava-odoo:
@robodoo override=ci/industry

PR #2021, opened 2026-04-21
[FIX] .*: fix non-stored fields in related field definitions

dhrs-odoo:
Before this commit we were using the non store field in the related field definitions but due to recent fix in standard* we are now unable to use those fields in the related definitions.

So in this commit we are replacing the related non store field reference as a computed field .

*https://github.com/odoo/odoo/commit/fe83265417217bb01b82fe777fb33a42650eb74b

### DIFF
--- a/art_craft/data/ir_model_fields.xml
+++ b/art_craft/data/ir_model_fields.xml
@@ -6,12 +6,4 @@
         <field name="ttype">boolean</field>
         <field name="field_description">Is consignee</field>
     </record>
-    <record id="field_stock_picking_x_is_it_consignee" model="ir.model.fields">
-        <field name="name">x_is_consignee</field>
-        <field name="model_id" ref="stock.model_stock_picking"/>
-        <field name="ttype">boolean</field>
-        <field name="readonly" eval="True"/>
-        <field name="related">purchase_id.x_is_consignee</field>
-        <field name="field_description">Is Consignee</field>
-    </record>
 </odoo>
--- a/bookstore/data/ir_model_fields.xml
+++ b/bookstore/data/ir_model_fields.xml
@@ -18,8 +18,12 @@ for template in self:
         <field name="name">x_last_sale_stock</field>
         <field name="field_description">Last Sale</field>
         <field name="ttype">date</field>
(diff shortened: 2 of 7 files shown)


art_craft/data/ir_model_fields.xml
```diff
@@ -11,7 +11,12 @@
...
         <field name="ttype">boolean</field>
         <field name="readonly" eval="True"/>
-        <field name="related">purchase_id.x_is_consignee</field>
+        <field name="depends">purchase_id.x_is_consignee</field>
         <field name="field_description">Is Consignee</field>
+        <field name="compute"><![CDATA[
+for picking in self:
+    picking['x_is_consignee'] = bool(picking.purchase_id and picking.purchase_id.x_is_consignee)
+]]></field>
+        <field name="store" eval="False"/>
```
vava-odoo: From what I see, this field is not used either. You can remove it and replace it by "purchase_id.x_is_consignee" in the domain and pre_domain.

bookstore/data/ir_model_fields.xml
```diff
@@ -18,8 +18,12 @@ for template in self:
         <field name="name">x_last_sale_stock</field>
         <field name="field_description">Last Sale</field>
         <field name="ttype">date</field>
-        <field name="related">product_tmpl_id.x_last_sale_date</field>
+        <field name="depends">product_tmpl_id.x_last_sale_date</field>
         <field name="model_id" ref="stock.model_stock_warehouse_orderpoint"/>
+        <field name="compute"><![CDATA[
+for orderpoint in self:
+    orderpoint['x_last_sale_stock'] = orderpoint.product_tmpl_id.x_last_sale_date if orderpoint.product_tmpl_id else False
```
vava-odoo: you probably don't need the if clause, do you?
```suggestion
    orderpoint['x_last_sale_stock'] = orderpoint.product_tmpl_id.x_last_sale_date
```

vava-odoo (approved):

vava-odoo (approved):
@robodoo r+

vava-odoo:
@robodoo squash

PR #2026, opened 2026-04-21
[FIX] booking_engine: fix cron not working after the first time

trcazier:
Before this commit, the `ir_cron_create_500_days_availability` cron could only work the first time. The code was correct when no availabilities were returned by the _read_group but incorrect when not empty. This commit fixes this and adds a test to validate that the cron can run twice.


vava-odoo (approved):
@robodoo r+

vava-odoo:
@robodoo override=ci/industry

PR #2029, opened 2026-04-21
[FIX] booking_engine: add test for unavailability due to calendar leave without resource

abdrahmanrashed:
This commit adds test to check you can't book rooms when  the rental period overlaps with`resource.calendar.leaves` with no `resource_id`


vava-odoo (approved):
@robodoo r+

PR #2062, opened 2026-04-27
[IMP] construction_developer: add WBS

trcazier:
This commit adds the Work Breakdown Structure functionality.
From a button on the SO, generate tasks in a project based on the content
of the SO to plan the work ahead of time.

task-5951093

### DIFF
--- a/construction/data/knowledge_article.xml
+++ b/construction/data/knowledge_article.xml
@@ -484,6 +484,62 @@
                 </div>
             </div>
         </div>
+
+        <h5>🧩 Work Breakdown Structure straight from your quote (Construction Developer)</h5>
+        <p>
+            Structuring a complex construction project into a clear <strong>Work Breakdown Structure</strong> (WBS) is essential for tracking progress and costs.
+            Instead of manual entry, Odoo automatically generates a hierarchical project structure directly from your Sales Order.
+        </p>
+        <p><strong>How to use it:</strong></p>
+        <ul>
+            <li>
+                Open a freshly created sale order organized with sections and possibly subsections.
+            </li>
+            <li>
+                From the gear icon (⚙️), fire the Create WBS action.
+            </li>
+            <li>
+                Reach your Gantt chart from the Tasks smart button or from the Project App, changing the task view to Gantt.
+            </li>
+            <li>
+                Expand the view with the double arrow button (↕️) to appreciate more details!
+            </li>
+        </ul>
+        <div
+            class="o_editor_banner user-select-none o-contenteditable-false lh-1 d-flex align-items-center alert alert-info pb-0 pt-3"
+            data-oe-role="status" contenteditable="false" role="status">
+            <i class="o_editor_banner_icon mb-3 fst-normal" data-oe-aria-label="Banner Info"
+                aria-label="Banner Info">💡</i>
+            <div class="o_editor_banner_content o-contenteditable-true w-100 px-3"
+                contenteditable="true">
+                <div class="o-paragraph">
+                    <p>
+                        Automated hierarchy:
+                    </p>
+                    <ul>
+                        <li>
+                            Sections and Subsections are converted into Parent Tasks.
+                        </li>
+                        <li>
+                            Product Sale Order Lines with Work Items are converted into Execution Tasks 
+                            (Work Packages) capturing the needed time based on components (assigned to 
+                            categories of Cost Nature Labour or Equipment).
+                        </li>
+                        <li>
+                            The Sale Order Line numbering (e.g., 1.4.2.) is applied to every task to 
+                            reflect its exact position in the project tree.
+                        </li>
+                    </ul>
+                    <p>
+                        Tasks are automatically sequenced in time (one task per day by default) starting
+                        from today, providing an instant baseline for your Gantt Chart.
+                    </p>
+                </div>
+            </div>
+        </div>
+
+
+
         <h5>🛒 Material Shopping List (Construction Developer)</h5>
         <p>
             Eliminate manual entry and guesswork by transforming your technical quotes into actionable procurement data.
@@ -497,7 +553,7 @@
                 Prepare your Sale Order with structured sections and articles containing Work Items.
             </li>
             <li>
-                Access your consolidated requirements via the Procurement smart button as well as from the <strong>Inventory App</strong> and the Project Top Bar <strong>"Needs" button</strong>
+                Access your consolidated requirements via the Procurement smart button as well as from the <strong>Inventory App</strong> and the Project Top Bar <strong>"Procurement" button</strong>
             </li>
         </ul>
         <div
--- a/construction_developer/data/base_automation.xml
+++ b/construction_developer/data/base_automation.xml
@@ -10,10 +10,33 @@
     </record>
     <record id="automation_initialize_sales_order_line_sequence_on_sales_order_create" model="base.automation">
         <field name="model_id" ref="sale.model_sale_order"/>
-        <field name="action_server_ids" eval="[(6, 0, [ref('action_initialize_sales_order_line_sequence_on_sales_order_create')])]"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_setup_sol_numbering'), ref('action_initialize_sales_order_line_sequence_on_sales_order_create')])]"/>
         <field name="trigger">on_create_or_write</field>
         <field name="name">Construction Developer: Initialize Sales Order Line Sequence on Sales Order create or write</field>
         <field name="trigger_field_ids" eval="[(6, 0, [ref('sale.field_sale_order__order_line')])]"/>
+    </record>
+    <record id="automation_update_sol_numbering_on_sol_change" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_update_sol_numbering')])]"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="name">Construction Developer: Update Sales Order Line Number on SOL write</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('sale.field_sale_order_line__sequence'), ref('sale.field_sale_order_line__parent_id')])]"/>
+    </record>
+    <record id="automation_update_and_remove_wbs_project_on_task_project_change" model="base.automation">
+        <field name="model_id" ref="project.model_project_task"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_update_and_remove_wbs_project_on_task_project_change')])]"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="filter_domain">['|', ('sale_line_id.order_id.x_wbs_project', '!=', False), ('x_sale_line_id.order_id.x_wbs_project', '!=', False)]</field>
+        <field name="name">Construction Developer: Update Analytic Account and Delete WBS Project on Task Project change</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('project.field_project_task__project_id')])]"/>
+    </record>
+    <record id="automation_sale_order_update_wbs_on_confirm" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_sale_order_update_wbs_on_confirm')])]"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="name">Construction Developer: Update WBS on Sale Order create or write</field>
+        <field name="filter_domain">[('state', '!=', 'draft'), ('x_wbs_project', '!=', False)]</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('sale.field_sale_order__state')])]"/>
     </record>
     <record id="automation_update_progress_increment_on_change" model="base.automation">
         <field name="name">Construction Developer: Update Sale Order Line Progress Increment on Quantity Increment change</field>
@@ -156,5 +179,12 @@
         <field name="filter_pre_domain">[("move_type", "=", "in_invoice")]</field>
         <field name="filter_domain">[('state', '=', 'posted')]</field>
         <field name="trigger_field_ids" eval="[(6, 0, [ref('account.field_account_move__state')])]"/>
+    </record>
+    <record id="automation_project_task_on_create_or_write" model="base.automation">
+        <field name="model_id" ref="project.model_project_task"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_update_parent_task_dates_and_time')])]"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="name">Construction Developer: Task - On task date &amp; time update</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('project.field_project_task__allocated_hours'), ref('project.field_project_task__date_deadline'), ref('project.field_project_task__parent_id'), ref('project_enterprise.field_project_task__planned_date_begin')])]"/>
     </record>
 </odoo>
--- a/construction_developer/data/ir_actions_server.xml
+++ b/construction_developer/data/ir_actions_server.xml
@@ -14,17 +14,55 @@ for record in records: record['x_cost_nature'] = record.x_cost_nature or record.
         <field name="name">Construction Developer: Initialize Sales Order Line Sequence on Sales Order create</field>
         <field name="state">code</field>
         <field name="code"><![CDATA[
-sequences = record.order_line.mapped('sequence')
-if len(sequences) != len(set(sequences)):  # not all different: create order
-    new_sols = []
-    for sol_i, sol in enumerate(record.order_line):
-        previous_same_sequences = sum(seq == sequences[sol_i] for seq in sequences[:sol_i + 1])
-        # Handle duplicates by adding fractional offsets which will be used to order the list
-        new_sols.append((sol.id, sol.sequence + (previous_same_sequences / sequences.count(sol.sequence))))
-    seqs = {sid: i for (i, (sid, _)) in enumerate(sorted(new_sols, key=lambda soltup: soltup[1]))}         # Associate each sid with its sequence (order) when sorted
-else:  # all different: use existing order
-    seqs = {sol.id: i for (i, sol) in enumerate(sorted(record.order_line, key=lambda sol: sol.sequence))}  # Associate each sid with its sequence (order) when sorted
-for sol in record.order_line: sol['sequence'] = seqs[sol.id] + 1
+for record in records:
+    sequences = record.order_line.mapped('sequence')
+    if len(sequences) != len(set(sequences)):  # not all different: create order
+        new_sols = []
+        for sol_i, sol in enumerate(record.order_line):
+            previous_same_sequences = sum(seq == sequences[sol_i] for seq in sequences[:sol_i + 1])
+            # Handle duplicates by adding fractional offsets which will be used to order the list
+            new_sols.append((sol.id, sol.sequence + (previous_same_sequences / sequences.count(sol.sequence))))
+        seqs = {sid: i for (i, (sid, _)) in enumerate(sorted(new_sols, key=lambda soltup: soltup[1]))}         # Associate each sid with its sequence (order) when sorted
+    else:  # all different: use existing order
+        seqs = {sol.id: i for (i, sol) in enumerate(sorted(record.order_line, key=lambda sol: sol.sequence))}  # Associate each sid with its sequence (order) when sorted
+    for sol in record.order_line: sol['sequence'] = seqs[sol.id] + 1
+]]></field>
+    </record>
+    <record id="server_action_update_sol_numbering" model="ir.actions.server">
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="state">code</field>
+        <field name="name">Construction Developer: Update SOL numbering</field>
+        <field name="code"><![CDATA[
+env.ref('construction_developer.server_action_setup_sol_numbering').with_context(active_ids = records.order_id.ids, active_model="sale.order").run()
+]]></field>
+    </record>
+    <record id="server_action_setup_sol_numbering" model="ir.actions.server">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Construction Developer: Setup SOL numbering</field>
+        <field name="code"><![CDATA[
+for record in records:
+    l1, l2, l3 = 0, 0, 0
+    
+    for sol in (sols := list(record.order_line.filtered(lambda sol: sol.display_type in [False, 'line_section', 'line_subsection']).sorted('sequence'))):
+        # Cannot use parent_id: when changing a line to another nesting level, the parent_id field
+        # is not updated yet when this automation is called
+        # Note: if a subsection is dragged out of a section and placed at the very top, it becomes a section
+        above_section = list(filter(lambda sol: sol.display_type != False, sols[:sols.index(sol)]))[-1:]
+
+        # top-level row (section, or product above first section)
+        if sol.display_type == 'line_section' or (sol.display_type == False and not above_section):
+            l1 += 1
+        # middle-level row (subsection, or product under a section)
+        elif sol.display_type == 'line_subsection' or above_section[0].display_type == 'line_section':
+            l2 += 1
+        # bottom-level row (product)
+        else:
+            l3 += 1
+        l2 = l2 * (sol.display_type != 'line_section'); l3 = l3 * (sol.display_type == False)
+
+        new_item_number = "".join([f"{n}." for n in [l1, l2, l3] if n])
+        sol.write({'x_item_number': new_item_number} if sol.x_item_number != new_item_number else {})
 ]]></field>
     </record>
     <record id="action_sol_update_progress_increment_on_change" model="ir.actions.server">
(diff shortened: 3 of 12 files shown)


construction/data/knowledge_article.xml
```diff
@@ -484,6 +484,62 @@
...
+
+        <h5>🧩 Work Breakdown Structure straight from your quote (Construction Developer)</h5>
+        <p>
+            Structuring a complex construction project into a clear <strong>Work Breakdown Structure</strong> (WBS) is essential for tracking progress and costs.
+            Instead of manual entry, Odoo automatically generates a hierarchical project structure directly from your Sales Order.
+        </p>
+        <p><strong>How to use it:</strong></p>
+        <ul>
+            <li>
+                Open a freshly created sale order organized with a sections and eventually subsections.
```
vava-odoo:
```suggestion
                Open a freshly created sale order organized with sections and possibly subsections.
```

construction_developer/data/ir_actions_act_window.xml
```diff
@@ -35,6 +35,16 @@
         <field name="target">new</field>
         <field name="context">{'default_x_is_template': False, 'default_x_targeted_so_id': context.get('so_id')}</field>
     </record>
+    <record id="action_window_wbs_project" model="ir.actions.act_window">
```
vava-odoo: I don't see where this is used

construction_developer/static/src/js/patch.js
```diff
@@ -30,3 +30,13 @@ patch(ListRenderer.prototype, { formatGroupAggregate(group, column) {
             aggregates["x_margin_percent"].rawValue = aggregates["x_total_price"].rawValue ? 100 * (aggregates["x_margin"].rawValue) / aggregates["x_total_price"].rawValue : 0;
             aggregates["x_margin_percent"].value = Math.round(aggregates["x_margin_percent"].rawValue).toString() + " %";}
         return aggregates;}})
+
+// In the Sale Order form view, in the SOL list view
+// Adds the Item number column to sections & subsections
+patch(SaleOrderLineListRenderer.prototype, { getSectionColumns(columns, record) {
```
vava-odoo, line 37: I don't see this method in standard sales modules 🤔 Not in 19, not in master 🤷

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+    <record id="server_action_update_sol_numbering" model="ir.actions.server">
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="state">code</field>
+        <field name="name">Construction Developer: Update SOL numbering</field>
+        <field name="code"><![CDATA[
+env.ref('construction_developer.server_action_setup_sol_numbering').with_context(active_ids = records.order_id.ids, active_model="sale.order").run()
+]]></field>
+    </record>
+    <record id="server_action_setup_sol_numbering" model="ir.actions.server">
+        <field name="model_id" ref="sale.model_sale_order"/>
```
vava-odoo: maybe group these two with the existing one on top?

construction_developer/data/ir_model_fields.xml
```diff
@@ -824,6 +848,84 @@ for record in self:
         <field name="relation">product.category</field>
         <field name="store" eval="True"/>
     </record>
+  <record id="field_project_task_x_full_time_equivalent" model="ir.model.fields">
+    <field name="ttype">float</field>
+    <field name="depends">planned_date_begin,date_deadline,allocated_hours</field>
+    <field name="field_description">Full Time Equivalent</field>
+    <field name="model_id" ref="project.model_project_task"/>
+    <field name="name">x_full_time_equivalent</field>
```
vava-odoo: Is it indicative? I don't think this needs to be stored

construction_developer/data/ir_model_fields.xml
```diff
@@ -824,6 +848,84 @@ for record in self:
...
+            theoretical_capacity = (days_in_period * 8.0) + min(8.0, remaining_hours)
+            
+            if theoretical_capacity > 0:
+                record['x_full_time_equivalent'] = record.allocated_hours / theoretical_capacity
+            else:
+                record['x_full_time_equivalent'] = 0.0
+        else:
+            record['x_full_time_equivalent'] = 0.0
+    else: 
+        record['x_full_time_equivalent'] = 0.0
```
vava-odoo: Why not put a `self['x_full_time_equivalent'] = 0.0` before the loop to avoid all those else clauses?

construction_developer/data/ir_model_fields.xml
```diff
@@ -824,6 +848,84 @@ for record in self:
...
+for record in self:
+    if record.planned_date_begin and record.date_deadline and record.allocated_hours > 0:
+        duration_seconds = (record.date_deadline - record.planned_date_begin).total_seconds()
+        if duration_seconds > 0:
+            days_in_period = int(duration_seconds // (24 * 3600))
+            
+            # Get the remaining seconds after extracting full days, convert to hours
+            remaining_hours = (duration_seconds % (24 * 3600)) / 3600.0
+            
+            theoretical_capacity = (days_in_period * 8.0) + min(8.0, remaining_hours)
```
vava-odoo, line 889: I guess you hardcode that a worker works 8 hours/day, 5 days a week. But that is not true at all for all cases (I know it's in the spec, but let's challenge this)

construction_developer/data/ir_model_fields.xml
```diff
@@ -824,6 +848,84 @@ for record in self:
...
+                record['x_full_time_equivalent'] = 0.0
+        else:
+            record['x_full_time_equivalent'] = 0.0
+    else: 
+        record['x_full_time_equivalent'] = 0.0
+]]></field>
+  </record>
+  <record id="field_sale_order_line_x_wbs_ready" model="ir.model.fields">
+    <field name="ttype">boolean</field>
+    <field name="copied" eval="True"/>
```
vava-odoo: I don't think this field should be copied

construction_developer/data/ir_model_fields.xml
```diff
@@ -824,6 +848,84 @@ for record in self:
...
+  </record>
+  <record id="field_sale_order_line_x_task_progress" model="ir.model.fields">
+    <field name="ttype">float</field>
+    <field name="related">task_id.progress</field>
+    <field name="field_description">Progress</field>
+    <field name="model_id" ref="sale.model_sale_order_line"/>
+    <field name="name">x_task_progress</field>
+    <field name="readonly" eval="True"/>
+    <field name="store" eval="False"/>
+  </record>
```
vava-odoo: Seems unused 
```suggestion
```

construction_developer/data/ir_model_fields.xml
```diff
@@ -824,6 +848,84 @@ for record in self:
...
+    <field name="ttype">one2many</field>
+    <field name="field_description">Sale Order Line Ids</field>
+    <field name="model_id" ref="sale.model_sale_order_line"/>
+    <field name="name">x_sale_order_line_ids</field>
+    <field name="relation">project.task</field>
+    <field name="relation_field">sale_line_id</field>
+  </record>
+  <record id="field_sale_order_line_x_work_progress" model="ir.model.fields">
+    <field name="ttype">float</field>
+    <field name="related">x_sale_order_line_ids.progress</field>
```
vava-odoo: Does that work with several lines? 🤔

construction_developer/data/ir_model_fields.xml
```diff
@@ -824,6 +848,84 @@ for record in self:
...
+    <field name="related">task_id.progress</field>
+    <field name="field_description">Progress</field>
+    <field name="model_id" ref="sale.model_sale_order_line"/>
+    <field name="name">x_task_progress</field>
+    <field name="readonly" eval="True"/>
+    <field name="store" eval="False"/>
+  </record>
+  <record id="field_sale_order_line_x_sale_order_line_ids" model="ir.model.fields">
+    <field name="ttype">one2many</field>
+    <field name="field_description">Sale Order Line Ids</field>
```
vava-odoo: This links to project tasks then, not sales order lines

construction_developer/demo/project_task_post.xml
```diff
@@ -0,0 +1,104 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+<!-- 
+    These tasks were created by the creation of the WBS.
+    Assigns them an ID to be used in other files.
```
vava-odoo, line 5: Ok if you use it in other files. But first occurrence (project_task_325) is only used here, so a simple call to write should be enough (no need for a xmlid)

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+    </record>
+
+    <record id="server_action_transfer_wbs" model="ir.actions.server">
+        <field name="model_id" ref="project.model_project_project"/>
+        <field name="state">code</field>
+        <field name="name">Construction Developer: Transfer WBS</field>
+        <field name="code"><![CDATA[
+wbss = records.filtered(lambda p: any([(t.x_sale_line_id or t.sale_line_id).order_id.x_wbs_project.id == p.id for t in p.tasks]))[:1]
+projects = records.filtered(lambda p: any([(t.x_sale_line_id or t.sale_line_id).order_id.x_wbs_project.id != p.id for t in p.tasks]))[:1]
+if len(records) != 2 or not wbss or not projects:
```
vava-odoo: tell me if I miss something, but isn't it `projects = records - wbss`?

On another hand, the slicing at the end shouldn't be necessary (if you expect 2 `records`, there should be one `wbss` and one `projects`)

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+    action = {'type': 'ir.actions.client', 'tag': 'display_notification', 'params': {'message': "Error: Can only Transfer one WBS project to one non-WBS project.", 'type': 'danger'}}
+else:
+    wbs_name, project_name, n = wbss[0].name, projects[0].name, len(wbss[0].tasks)
+    wbss[0].tasks.write({'project_id': projects[0].id})
```
vava-odoo: same remark about the slicing here

construction_developer/data/ir_ui_view.xml
```diff
@@ -298,6 +337,19 @@
...
                     </div>
                 </button>
+                <button
+                    name="%(server_action_show_wbs)d"
+                    type="action"
+                    class="oe_stat_button"
+                    icon="fa-tasks"
+                    invisible="not x_wbs_project"
+                    groups="project.group_project_user"
+                    context="{'active_id': id}"
```
vava-odoo: I'm surprised you manually need to add it here

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+]]></field>
+    </record>
+
+    <record id="server_action_show_wbs" model="ir.actions.server">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Construction Developer: Create WBS</field>
+        <field name="code"><![CDATA[
+action = env['ir.actions.actions'].with_context(active_id=(p_id := record.x_wbs_project.id))._for_xml_id('project.act_project_project_2_project_task_all')
+action['context'] = {'active_id': p_id, 'default_project_id': p_id, 'default_user_ids': [env.uid]}
```
vava-odoo, line 449: why you need to define active_id twice?

construction_developer/data/ir_ui_view.xml
```diff
@@ -743,4 +795,81 @@
...
+        <field name="inherit_id" ref="project.view_task_form2"/>
+        <field name="mode">extension</field>
+        <field name="model">project.task</field>
+        <field name="name">project.task.view.form.inherit</field>
+        <field name="priority">9990</field>
+        <field name="type">form</field>
+        <field name="active" eval="True"/>
+        <field name="arch" type="xml">
+            <data>
+                <xpath expr="/form/sheet/group/group[2]/div[@id='allocated_hours_container']" position="after">
```
vava-odoo: I bet this would be more stable over time if we reduce the path (o be tested)
```suggestion
                <xpath expr="//div[@id='allocated_hours_container']" position="after">
```

construction_developer/data/ir_ui_view.xml
```diff
@@ -743,4 +795,81 @@
...
+                </xpath>
+            </data>
+        </field>
+    </record>
+    <record id="view_sale_project_inherit_form_inherit" model="ir.ui.view">
+    <field name="inherit_id" ref="sale_project.view_sale_project_inherit_form"/>
+    <field name="mode">extension</field>
+    <field name="model">project.task</field>
+    <field name="name">project.task.view.form.inherit</field>
+    <field name="priority">9990</field>
```
vava-odoo, line 821: How much do you want?
Yes.

construction_developer/data/ir_ui_view.xml
```diff
@@ -743,4 +795,81 @@
...
+    <record id="view_task_form2_inherit" model="ir.ui.view">
+        <field name="inherit_id" ref="project.view_task_form2"/>
+        <field name="mode">extension</field>
+        <field name="model">project.task</field>
+        <field name="name">project.task.view.form.inherit</field>
+        <field name="priority">9990</field>
+        <field name="type">form</field>
+        <field name="active" eval="True"/>
+        <field name="arch" type="xml">
+            <data>
```
vava-odoo: useless here
```suggestion
```

construction_developer/demo/project_task_post.xml
```diff
@@ -0,0 +1,104 @@
...
+-->
+
+    <!-- S00003: Ground (section) -->
+    <function model="ir.model.data" name="_update_xmlids">
+        <value model="project.task" eval="[{
+            'xml_id': 'construction_developer.project_task_325', 
+            'record': obj().search([('sale_line_id', '=', ref('construction_developer.sale_order_line_325'))])}]"/>
```
vava-odoo: probably need to add a limit=1

construction_developer/demo/project_task_post.xml
```diff
@@ -0,0 +1,104 @@
...
+    <function model="ir.model.data" name="_update_xmlids">
+        <value model="project.task" eval="[{
+            'xml_id': 'construction_developer.project_task_319', 
+            'record': obj().search([('sale_line_id', '=', ref('construction_developer.sale_order_line_319'))])}]"/>
+    </function>
+    <record id="project_task_319" model="project.task">
+        <field name="stage_id" ref="construction.planning_project_stage_3" />
+        <field name="state">1_done</field>
+        <field name="planned_date_start" eval="datetime.today().date() - relativedelta(days=3)"/>
+        <field name="date_deadline" eval="datetime.today().date() - relativedelta(days=3)"/>
```
vava-odoo, line 29: same date? 🤔

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+        # Cannot use parent_id: when changing a line to another nesting level, the parent_id field
+        # is not updated yet when this automation is called
+        # Note: if a subsection is dragged out of a section and placed at the very top, it becomes a section
+        above_section = list(filter(lambda sol: sol.display_type != False, sols[:sols.index(sol)]))[-1:]
+
+        # top-level row (section, or product above first section)
+        if sol.display_type == 'line_section' or (sol.display_type == False and not above_section):
+            l1 += 1
+        # middle-level row (subsection, or product under a section)
+        elif sol.display_type == 'line_subsection' or above_section[0].display_type == 'line_section':
```
vava-odoo: are you sure above_section does always exist here? maybe this?
```suggestion
        elif sol.display_type == 'line_subsection' or above_section[:1].display_type == 'line_section':
```

construction_developer/data/base_automation.xml
```diff
@@ -10,11 +10,33 @@
...
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_update_sol_numbering')])]"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="name">Construction Developer: Update Sales Order Line Number on SOL write</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('sale.field_sale_order_line__sequence'), ref('sale.field_sale_order_line__parent_id')])]"/>
+    </record>
+    <record id="automation_update_and_remove_wbs_project_on_task_project_change" model="base.automation">
+        <field name="model_id" ref="project.model_project_task"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_update_and_remove_wbs_project_on_task_project_change')])]"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="filter_domain">['|', ('sale_line_id', '!=', False), ('x_sale_line_id', '!=', False), '|', ('sale_line_id.order_id.x_wbs_project', '!=', False), ('x_sale_line_id.order_id.x_wbs_project', '!=', False)]</field>
```
vava-odoo: is the first part needed? I mean, `sale_line_id.order_id.x_wbs_project` can not be truthy if there is no `sale_line_id`
```suggestion
        <field name="filter_domain">['|', ('sale_line_id.order_id.x_wbs_project', '!=', False), ('x_sale_line_id.order_id.x_wbs_project', '!=', False)]</field>
```

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+]]></field>
+    </record>
+
+    <record id="server_action_update_and_remove_wbs_project_on_task_project_change" model="ir.actions.server">
+        <field name="model_id" ref="project.model_project_task"/>
+        <field name="state">code</field>
+        <field name="name">Construction Developer: Update Analytic Account and Delete WBS Project on Task Project change</field>
+        <field name="code"><![CDATA[
+for record in records:
+    sol = record.sale_line_id or record.x_sale_line_id
```
vava-odoo, line 480: in other place, we give priority to `x_sale_line_id`. So why change this here?

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+    project = (so := sol.order_id).x_wbs_project
+    if project and not env['project.task'].search_count([('project_id', '=', project.id)]):
```
vava-odoo:
```suggestion
    if project and not env['project.task'].search_count([('project_id', '=', project.id)], limit=1):
```

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+]]></field>
+    </record>
+
+    <record id="server_action_sale_order_update_wbs_on_confirm" model="ir.actions.server">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Construction Developer: Update WBS on Sale Order confirm</field>
+        <field name="code"><![CDATA[
+for record in records:
+    if record.state != 'draft' and (project := record.x_wbs_project):
```
vava-odoo: Shouldn't this be in the filter_domain?

construction_developer/data/base_automation.xml
```diff
@@ -10,11 +10,33 @@
...
+        <field name="name">Construction Developer: Update Analytic Account and Delete WBS Project on Task Project change</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('project.field_project_task__project_id')])]"/>
+    </record>
+    <record id="automation_sale_order_update_wbs_on_confirm" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_sale_order_update_wbs_on_confirm')])]"/>
+        <field name="trigger">on_create_or_write</field>
```
vava-odoo, line 36: shouldn't it be 'on stage changed'?

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+        if (wbs_tasks := project.tasks.filtered('x_sale_line_id')):
+            for task in wbs_tasks:
+                task.write({'x_sale_line_id': False, 'sale_line_id': task.x_sale_line_id.id})
+
+            so_id = wbs_tasks.filtered('sale_line_id')[0].sale_line_id.order_id.id
```
vava-odoo: why do you need to filter again? You know it exists for all, since you moved it from x_sale_line_id (that exists)
```suggestion
            so_id = wbs_tasks[0].sale_line_id.order_id.id
```

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+]]></field>
+    </record>
+
+    <record id="server_action_create_wbs" model="ir.actions.server">
+        <field name="binding_model_id" ref="sale.model_sale_order"/>
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Construction Developer: Create WBS</field>
+        <field name="code"><![CDATA[
+uom_hour = env['uom.uom'].browse(4)
```
vava-odoo: 4?! 🤌 
You need to refer to an xmlid

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+
+for record in records:
+    # when a SO is confirmed: use the default field sale_line_id
+    # when a SO is NOT confirmed, use x_sale_line_id otherwise the SO is automatically confirmed
+    
+    line_field = 'sale_line_id' if (confirmed := record.state != 'draft') else 'x_sale_line_id'
```
vava-odoo, line 513: what about canceled orders?

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+    if not record.project_id:
+        record['project_id'] = project.id
+
+    record['x_wbs_project'] = project.id
+
+    # Create default stages for new projects
+    if not project.type_ids:
+        # By default: takes the default stages defined in the construction module
+        default_stage_ids = [env.ref(f'construction.planning_project_stage_{id}', raise_if_not_found=False).id for id in [0,1,2,3,4]]
+        default_stages = env['project.task.type'].search([('id', 'in', default_stage_ids)]) if default_stage_ids else False
```
vava-odoo: Even if you use `raise_if_not_found=False`, if there is no ref, `.id` will fail
```bash
>>> self.env.ref('this_does_not_exist', raise_if_not_found=False).id
Traceback (most recent call last):
  File "/usr/lib/python3.12/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'id'
```
but I wonder if browse wouldn't be smarter
```suggestion
        default_stage_ids = [env.ref(f'construction.planning_project_stage_{id}', raise_if_not_found=False) for id in [0,1,2,3,4]]
        default_stages = env['project.task.type'].browse([t.id for t in default_stage_ids if t])
```

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+
+        # If they don't exist, create a new default stage
+        if not default_stages:
```
vava-odoo: so if one exists, you keep that one, but if none exists, you create 5. Weird, no?
Shouldn't we copy the stages from another project?
I'm a bit lost in the logic here

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+            default_stages = env['project.task.type'].create([
+                {'name': name, 'sequence': sequence, 'fold': fold, 'project_ids': False, 'color': color}
+                for name, sequence, fold, color in [
+                    (env._('New'), 1, False, 2),
```
vava-odoo: not sure the translation tool will work here 🤔 if you export the pot, do you see this. I bet not

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +390,289 @@ if so_ids != [0]:
...
+                    (env._('Planned'), 5, False, 3),
+                    (env._('In Progress'), 10, False, 4),
+                    (env._('Done'), 20, True, 10),
+                    (env._('Cancelled'), 25, True, False),
+            ]])
+        project['type_ids'] = default_stages.ids
+
+    # WBS CREATION & UPDATE LOOP
+    parent_lvl1, parent_lvl2 = False, False
```
vava-odoo, line 536: I stop here ... still need to review this one and the next... But time to give you some feedback already

construction_developer/static/src/js/patch.js
trcazier: https://github.com/odoo/odoo/blob/459a775a066fe1465e53fc751045baed39f80118/addons/sale/static/src/js/sale_order_line_field/sale_order_line_field.js#L64

https://github.com/odoo/odoo/blob/459a775a066fe1465e53fc751045baed39f80118/addons/account/static/src/components/product_label_section_and_note_field/product_label_section_and_note_field_o2m.js#L10

https://github.com/odoo/odoo/blob/459a775a066fe1465e53fc751045baed39f80118/addons/resource/static/src/section_list_renderer.js#L37

construction_developer/data/ir_model_fields.xml
trcazier: Because then it will be 0 instead of False when it can't be computed

construction_developer/data/ir_model_fields.xml
trcazier: How could we do to change this?

construction_developer/data/ir_actions_server.xml
trcazier: This also fixes a bug that would have arisen in the case of transferring to a project with no tasks. Thanks

construction_developer/data/ir_actions_server.xml
trcazier: It does not get picked up otherwise, probably because it's in the context for the function that's called but not the one constructed for the return?

construction_developer/demo/project_task_post.xml
trcazier: Yes, this makes it 1 day

construction_developer/data/ir_actions_server.xml
trcazier: How can we be sure when it's a product that we have an above section?
Because if it doesn't, it's a top row product and caught in the second part of the first if:
`if sol.display_type == 'line_section' or (sol.display_type == False and not above_section)`

construction_developer/data/ir_actions_server.xml
trcazier: It doesn't really matter, does it? `x_sale_line_id` is only used when the sale order is not confirmed, and gets transferred to `sale_line_id` when it is. I don't think it matters the order we access them, we just have to find which one it's in, right?

construction_developer/data/ir_actions_server.xml
trcazier: Limit on a search_count...?

construction_developer/data/base_automation.xml
trcazier: Technically yes but that's not really what we want to do, we just want to transfer the x_ field to the regular field once the SO isn't a draft anymore which happens only once (when it is confirmed, not when its stage changes further)

construction_developer/data/ir_actions_server.xml
trcazier: You mean you think we should skip the creation of the WBS if the order is canceled?

construction_developer/data/ir_actions_server.xml
trcazier: If we were to copy the stages from another project, which one should we choose? The latest one? A demo project? The project template? What if the user deleted it? We still have to provide a stage, or the tasks will be floating in the project without a stage
So by default we use the stages we provided, and if the user deleted them, well, we can't randomly pick one of them and hope it's the correct one, so we create them ourselves
I asked SAVP about creating only a 'New' or all of them, and he chose all of them and said that at this point it's an edge case
What do you think?

construction_developer/data/ir_actions_server.xml
trcazier: I don't see it, indeed. Any idea what we could do to solve this?

construction_developer/data/ir_actions_server.xml
trcazier: Thank you :smile:

construction_developer/data/ir_model_fields.xml
```diff
@@ -824,6 +848,78 @@ for record in self:
...
+  <record id="field_sale_order_line_x_work_progress" model="ir.model.fields">
+    <field name="ttype">float</field>
+    <field name="field_description">Work Progress</field>
+    <field name="model_id" ref="sale.model_sale_order_line"/>
+    <field name="name">x_work_progress</field>
+    <field name="readonly" eval="True"/>
+    <field name="store" eval="False"/>
+    <field name="depends">x_task_ids,x_task_ids.progress</field>
+    <field name="compute">
+for sol in self: sol['x_work_progress'] = sum(sol.x_task_ids.mapped('progress')) / len(sol.x_task_ids)
```
vava-odoo: probably safer with an if clause in case no task is linked

construction_developer/data/ir_model_fields.xml
vava-odoo: Is it a real problem?
I mean, the ORM treats 0 as Falsy anyway. It would just display 0 instead of empty. And assigning False will be converted to 0. 
I don't see the issue.

construction_developer/data/ir_model_fields.xml
vava-odoo: description is still wrong, no?

vava-odoo: by the way, isn't it costly to add a one2many just to ease the compute method?
wouldn't be best to search (and compute) on the fly?
but then, there is no dependency field (though that should not an issue)

construction_developer/data/ir_model_fields.xml
```diff
@@ -789,6 +797,22 @@ for record in self:
...
+        <field name="model_id" ref="x_work_item_model"/>
+        <field name="relation">project.task</field>
+        <field name="store" eval="True"/>
+    </record>
+    <record id="field_x_work_item_line_x_task_start" model="ir.model.fields">
+        <field name="name">x_task_start</field>
+        <field name="field_description">Task Start Date</field>
+        <field name="ttype">datetime</field>
+        <field name="model_id" ref="x_work_item_line_model"/>
+        <field name="store" eval="True"/>
```
vava-odoo, line 832: if this is only for the group_by, no need to store

vava-odoo: oh no, there is a filter 🫠

construction_developer/data/ir_ui_view.xml
```diff
@@ -743,4 +794,79 @@
...
+    <field name="model">project.task</field>
+    <field name="name">project.task.view.gantt.inherit</field>
+    <field name="priority">99</field>
+    <field name="type">gantt</field>
+    <field name="active" eval="True"/>
+    <field name="arch" type="xml">
+      <data>
+        <xpath expr="/gantt" position="attributes">
+          <attribute name="default_group_by">parent_id</attribute>
+          <attribute name="display_mode">sparse</attribute>
```
vava-odoo, line 870: are we ok to enforce this on all task views? Or should we restrict this?

construction_developer/data/ir_actions_server.xml
vava-odoo: why not? any perf optimization is good to take, no?

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +427,249 @@ if so_ids != [0]:
...
+    <record id="server_action_update_and_remove_wbs_project_on_task_project_change" model="ir.actions.server">
+        <field name="model_id" ref="project.model_project_task"/>
+        <field name="state">code</field>
+        <field name="name">Construction Developer: Update Analytic Account and Delete WBS Project on Task Project change</field>
+        <field name="code"><![CDATA[
+for record in records:
+    sol = record.sale_line_id or record.x_sale_line_id
+    project = (so := sol.order_id).x_wbs_project
+    if project and not env['project.task'].search_count([('project_id', '=', project.id)]):
+        project['account_id'] = False
```
vava-odoo, line 483: do you need to edit the field if you're about to unlink it?

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +427,249 @@ if so_ids != [0]:
...
+        so['x_wbs_project'] = False
```
vava-odoo: and this is probably the default behaviour: if you remove the foreign key, this gets empty/falsy. No?

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +427,249 @@ if so_ids != [0]:
...
+    </record>
+
+    <record id="server_action_sale_order_update_wbs_on_confirm" model="ir.actions.server">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Construction Developer: Update WBS on Sale Order confirm</field>
+        <field name="code"><![CDATA[
+for record in records:
+    if (wbs_tasks := record.x_wbs_project.tasks.filtered('x_sale_line_id')):
+        for task in wbs_tasks:
```
vava-odoo: Avoid second loop
```suggestion
    for task in record.x_wbs_project.tasks:
         if task.x_sale_line_id:
```

construction_developer/data/ir_actions_server.xml
vava-odoo: this seems obvious from this far but 🤷

construction_developer/data/ir_actions_server.xml
```diff
@@ -389,6 +427,249 @@ if so_ids != [0]:
...
+    if not project:
+        project = env['project.project'].create({
+            'name': env._(f"{record.name} WBS - {record.partner_id.name}"),
+            'partner_id': record.partner_id.id,
+            'company_id': record.company_id.id,
+            'allow_timesheets': True,
+            'allow_billable': True,
+        } | ({} if not confirmed else {'sale_order_id': record.id}))  # If SO already confirmed but no project exists, we need to link the SO
+    if not record.project_id:
+        record['project_id'] = project.id
```
vava-odoo: so you always assign the project on the so but not always the so on the project 🤔

construction_developer/data/ir_actions_server.xml
vava-odoo: well yes. What happens in standard if you don't have defined stages or deleted existing ones? I'd look for the standard way and apply same idea here

construction_developer/data/ir_actions_server.xml
vava-odoo: I don't think there is a way for this to get translated. 
Try this script in French db and see the result. But I bet env._ has no power here, even more if the term is not in the pot

vava-odoo (commented):
Second review, and still not digged into the two big boys. Next time 🤞

construction_developer/data/ir_actions_server.xml
trcazier: > What happens in standard if you don't have defined stages or deleted existing ones?

When doing what?
In standard no stages are created and the tasks are set in a "None" folded stage when you assign them to a project with no stages, which is what happens here if we don't create stages

construction_developer/data/ir_actions_server.xml
trcazier: Yes because you can't delete a project that has an account_id linked to existing tasks

construction_developer/data/ir_actions_server.xml
trcazier: Indeed, it doesn't work.
What is also problematic is that it's not the only thing that seems to not get translated correctly
image

construction_developer/data/ir_model_fields.xml
vava-odoo: discussed with Pierre, let's keep it simple :+1:

construction_developer/static/src/js/patch.js
trcazier: My mistake, the third link should be this

https://github.com/odoo/odoo/blob/459a775a066fe1465e53fc751045baed39f80118/addons/account/static/src/components/section_and_note_fields_backend/section_and_note_fields_backend.js#L440

construction_developer/data/ir_actions_server.xml
```diff
@@ -401,4 +439,250 @@ if so_ids != [0]:
...
+for record in records:
+    # Query the DB directly to get the pre-changes values
+    env.cr.execute("SELECT planned_date_begin, date_deadline FROM project_task WHERE id = %s", (record.id,))
+    res = env.cr.fetchone()
+    old_start_raw, old_end_raw = res[0:2] if res else (False, False)
+    old_start = old_start_raw if isinstance(old_start_raw, datetime.datetime) else False
+    old_end = old_end_raw if isinstance(old_end_raw, datetime.datetime) else False
+    new_start, new_end = record.planned_date_begin, record.date_deadline
+
+    if not env.context.get('skip_task_date_recomputation'):
```
vava-odoo: couldn't you move this check at the beginning of the loop?

vava-odoo (commented):
👍

construction_developer/data/ir_actions_server.xml
vava-odoo: so let's remove all _ no?

construction_developer/data/ir_actions_server.xml
vava-odoo: I'd do the same, but :shrug:

vava-odoo:
please squash your commit :-)

vava-odoo (approved):
just need to add a setting for the loc

vava-odoo:
cancelled :melting_face:

PR #2080, opened 2026-04-29
[FIX] condominium: use record ID instead of record in search domain

abdrahmanrashed:
This commit sets the search domain in  the `ir_act_server_split_per_property` action to use record ID instead of the record itself.


vava-odoo (approved):
@robodoo r+

abdrahmanrashed:
@robodoo r+

vava-odoo:
@robodoo r+

PR #2085, opened 2026-04-29
[FIX] construction_developer: fix grouped list pager showing wrong total

UemuS:
The _postprocessReadGroup patch was unconditionally returning res_groups.length (the number of groups in the current page, e.g. 10) as the total group count, instead of the length value returned by the server (e.g. 148). This caused the pager to display "1-10/10" or "1-80/80" with no way to navigate to further pages, even when hundreds of groups existed.

Moves the early return inside the x_remark/x_stage_id branch where it belongs, and falls through to return the correct server-provided length for all other models.

Also adds missing sale_timesheet_enterprise dependency in construction module for the invoiced_timesheet field.

### DIFF
--- a/construction/__manifest__.py
+++ b/construction/__manifest__.py
@@ -13,6 +13,7 @@
         'sale_margin',
         'sale_project_forecast',
         'sale_timesheet',
+        'sale_timesheet_enterprise',
     ],
     'data': [
         'data/documents_folder.xml',
--- a/construction_developer/static/src/js/patch.js
+++ b/construction_developer/static/src/js/patch.js
@@ -10,8 +10,9 @@ patch(RelationalModel.prototype, { async _postprocessReadGroup(config, { groups,
     if (config.resModel == "x_remark" && groups.length > 0  && res_groups.length > 0 && (config.groupBy.includes("x_stage_id")) && groups[0].x_stage_id) {
         const stageIdsProjectsList = (await this.env.services.orm.searchRead("x_remark_stage", [], ['id', 'x_project_ids', 'x_sequence']));
         res_groups = res_groups.filter((group) => stageIdsProjectsList.some((stage) => stage.id === group.rawValue[0] && stage.x_project_ids.includes(config.context.default_x_project_id)))
-                               .map((group) => { group.x_sequence = stageIdsProjectsList.find((stage) => stage.id === group.rawValue[0]).x_sequence; return group; }).sort((group_a, group_b) => group_a.x_sequence - group_b.x_sequence);}
-    return { groups: res_groups, length: res_groups.length };}});
+                               .map((group) => { group.x_sequence = stageIdsProjectsList.find((stage) => stage.id === group.rawValue[0]).x_sequence; return group; }).sort((group_a, group_b) => group_a.x_sequence - group_b.x_sequence);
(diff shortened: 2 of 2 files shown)


UemuS:
Test failing seems unrelated..

construction/__manifest__.py
```diff
@@ -13,6 +13,7 @@
         'sale_margin',
         'sale_project_forecast',
         'sale_timesheet',
+        'sale_timesheet_enterprise',
```
vava-odoo, line 16: well spotted. But no need to specify both.
```suggestion
        'sale_timesheet_enterprise',
```

construction_developer/static/src/js/patch.js
```diff
@@ -10,8 +10,9 @@ patch(RelationalModel.prototype, { async _postprocessReadGroup(config, { groups,
     if (config.resModel == "x_remark" && groups.length > 0  && res_groups.length > 0 && (config.groupBy.includes("x_stage_id")) && groups[0].x_stage_id) {
         const stageIdsProjectsList = (await this.env.services.orm.searchRead("x_remark_stage", [], ['id', 'x_project_ids', 'x_sequence']));
         res_groups = res_groups.filter((group) => stageIdsProjectsList.some((stage) => stage.id === group.rawValue[0] && stage.x_project_ids.includes(config.context.default_x_project_id)))
-                               .map((group) => { group.x_sequence = stageIdsProjectsList.find((stage) => stage.id === group.rawValue[0]).x_sequence; return group; }).sort((group_a, group_b) => group_a.x_sequence - group_b.x_sequence);}
-    return { groups: res_groups, length: res_groups.length };}});
+                               .map((group) => { group.x_sequence = stageIdsProjectsList.find((stage) => stage.id === group.rawValue[0]).x_sequence; return group; }).sort((group_a, group_b) => group_a.x_sequence - group_b.x_sequence);
+        return { groups: res_groups, length: res_groups.length };}
```
vava-odoo, line 14: @trcazier could you check this?

vava-odoo (commented):
Thanks for the fix.
Isn't there any linked ticket?
Could you bump the minor version of both modules?
Isn't the issue impacting 19 as well?
Cheers

vava-odoo:
> Test failing seems unrelated..

yes you can ignore @UemuS it was only fixed earlier today https://github.com/odoo/industry/commit/027e2943c399126565f05a2a744e13198d9e967c
but indeed not related to your PR

construction_developer/static/src/js/patch.js
trcazier: Looks good to me! :+1:

UemuS:
@vava-odoo Hello, you're right the issue is impacting 19.0 as well (reproduced on a fresh empty DB)
I added reproduction steps on the new PR:
https://github.com/odoo/industry/pull/2088

trcazier (approved):
LGTM 🚀 
The code is the same from 19.0 to master, so the fix should be backported and forwardported indeed

vava-odoo:
Thank you both, let's continue on the other PR then https://github.com/odoo/industry/pull/2088
(next time, it's better to (put in draft and) retarget to another version instead ;-)

UemuS:
My bad, forgot I could simply rename the branch :')

PR #2088, opened 2026-04-30
[FIX] construction_developer: fix grouped list pager showing wrong total

UemuS:
ticket-6119505

The _postprocessReadGroup patch was unconditionally returning res_groups.length (the current page size) as the total group count, instead of the length value returned by the server. This caused the pager to display "1-10/10" or "1-80/80" with no way to navigate to further pages, even when hundreds of groups existed.

Moves the early return inside the x_remark/x_stage_id branch where it belongs, and falls through to return the correct server-provided length for all other models.

Also adds missing sale_timesheet_enterprise dependency in construction module for the invoiced_timesheet field.

Steps to reproduce:
- Install construction_developer on a fresh DB
- Create more than 10 invoices (for example, the issue happens on any grouping and any module)
- Open a grouped list view with more than 10 groups (e.g. group by invoice name)
- The pager shows "1-10/10" with no way to navigate to further pages

### DIFF
--- a/construction/__manifest__.py
+++ b/construction/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Construction Builder',
-    'version': '1.2',
+    'version': '1.3',
     'category': 'Construction',
     'depends': [
         'base_industry_data',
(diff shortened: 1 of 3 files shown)


vava-odoo (approved):
I just added a version bump in manifest. Otherwise, fine as is 👌 
Many thanks!
@robodoo r+

PR #2110, opened 2026-05-06
[FIX] booking_engine: prevent duplicate slot creation and ensure SOL linkage

dhrs-odoo:
Before this PR we replaced the `planning_slot_ids` (one2many) field in our 
code which was in relation with `sale_line_id` field due to which slots were 
created twice and also `sale_line_id` was not linked to the slot.

**Cause:**
- Since planning_slot_ids field is not in the form view of rental order, the 
  form view cannot set the context in the SOL created by the context that's why 
  the shift is no longer related to the SOL and the system is looking for an 
  available resource instead of setting the slot in which we create that rental 
  order.

**Fix:**
- We removed the code which was removing/replacing the planning_slot_ids field 
  from the view to ensure proper context propogation and avoid duplicate slot 
creation.

**opw - 5958916**

### DIFF
--- a/booking_engine/__manifest__.py
+++ b/booking_engine/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Booking Engine',
-    'version': '1.11',
+    'version': '1.12',
     'category': 'Hidden/Tools',
     'author': 'Odoo S.A.',
     'depends': [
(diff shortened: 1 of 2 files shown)


Pierre-savp:
🟢 to merge asap @vava-odoo

vava-odoo (commented):
@robodoo r+

PR #2112, opened 2026-05-06
[IMP] wine_merchant: add excise management to knowledge

abdrahmanrashed:
This commit adds a secton for excise management in the wine shop industry's knowledge article.

task-6107334


vava-odoo (approved):
@robodoo r+

PR #2117, opened 2026-05-07
[IMP] agri_equipment_rental: Added context to prevent mails

amas-odoo:
- Added context to prevent mails from generating.


vava-odoo (approved):
@robodoo r+

PR #2061, opened 2026-04-27
[IMP] booking_engine: add house keeping task assignment

abdrahmanrashed:
- Reworked `booking_engine` planning roles, adding new role category field to provide scalbility in handling different resource types.
- Replaced stayover and checkout cleaning flags with points to weight how much *effort* each stay offer would require to complete.
- Created new action and cron to automatically assign house keeping employees to house keeping tasks based on the cleaning points.

task-5929205

PR #2131, opened 2026-05-11
[ADD] .*: improve rental booking and housekeeping management

dhrs-odoo:
- add rental status indicators to quickly identify bookings for today's check-in, ongoing stays, and today's check-out directly from the view
- remove the default today filter
- add "Today's Check-In" and "Today's Check-Out" filters in the gantt view
- add default group by stage on the housekeeping board
- hide Guests tab and Tax button when no stay offer products are present in the SOL
- add many2one_avatar widget on x_guest_partner_id field
- hide the housekeeping section when the resource role is not linked to a stay offer product
- update demo data so orders with a start date before today and an end date after today are automatically checked in
- update demo data so orders with an end date before today are automatically checked out

Task-6144429

### DIFF
--- a/booking_engine/data/ir_model_fields.xml
+++ b/booking_engine/data/ir_model_fields.xml
@@ -789,4 +789,52 @@ Note that saving a new survey with this set will archive previous survey with th
       <field name="model_id" ref="sale.model_sale_order"/>
       <field name="ttype">boolean</field>
   </record>
+  <record id="x_booking_color_field" model="ir.model.fields">
+    <field name="name">x_booking_color</field>
+    <field name="field_description">Booking Color</field>
+    <field name="model_id" ref="planning.model_planning_slot"/>
+    <field name="ttype">integer</field>
+    <field name="store" eval="False"/>
+    <field name="depends">sale_order_id.rental_status,sale_order_id.rental_start_date,sale_order_id.rental_return_date,sale_order_id.is_late</field>
+    <field name="compute"><![CDATA[
+today_start = datetime.datetime.combine(datetime.date.today(), datetime.time.min)
+today_end = datetime.datetime.combine(datetime.date.today(), datetime.time.max)
+
+for slot in self:
+    order = slot.sale_order_id
+    if not order:
+        slot['x_booking_color'] = 5   # purple - internal
+    elif order.rental_status == 'pickup' and order.rental_start_date and today_start <= order.rental_start_date <= today_end:
+        slot['x_booking_color'] = 3   # yellow - check in due
+    elif order.rental_status == 'return' and order.rental_return_date and order.rental_return_date > today_end:
+        slot['x_booking_color'] = 10  # green - ongoing
+    elif order.rental_status == 'return' and order.rental_return_date and today_start <= order.rental_return_date <= today_end:
+        slot['x_booking_color'] = 4   # blue - check out due
+    elif (order.rental_status == 'pickup' and order.is_late) or (order.rental_status == 'return' and order.is_late):
+        slot['x_booking_color'] = 1   # red - attention
+    else:
+        slot['x_booking_color'] = False # grey - no focus
+      ]]></field>
+  </record>
+  <record id="x_rental_status_field" model="ir.model.fields">
+    <field name="ttype">selection</field>
+    <field name="field_description">Rental Status</field>
+    <field name="model_id" ref="planning.model_planning_slot"/>
+    <field name="name">x_rental_status</field>
+    <field name="related">sale_order_id.rental_status</field>
+    <field name="store">False</field>
+    <field name="readonly" eval="True"/>
+  </record>
+  <record id="x_has_room_offer_role_field" model="ir.model.fields">
+      <field name="name">x_has_room_offer_role</field>
+      <field name="field_description">Has Room Offer Role</field>
+      <field name="model_id" ref="resource.model_resource_resource"/>
+      <field name="ttype">boolean</field>
+      <field name="store" eval="False"/>
+      <field name="depends">role_ids.x_is_a_room_offer</field>
+      <field name="compute"><![CDATA[
+for record in self:
+    record['x_has_room_offer_role'] = any(role.x_is_a_room_offer for role in record.role_ids)
+  ]]></field>
+  </record>
 </odoo>
--- a/booking_engine/data/ir_ui_view.xml
+++ b/booking_engine/data/ir_ui_view.xml
@@ -54,14 +54,14 @@
           <attribute name="string">Check Out</attribute>
         </xpath>
         <xpath expr="//field[@name='is_rental_order']" position="after">
-          <button string="City Tax" type="action" name="%(open_x_city_tax_action)d" context="{'search_default_x_sale_order_id': id, 'default_x_sale_order_id': id}"/>
+          <button string="City Tax" type="action" name="%(open_x_city_tax_action)d" context="{'search_default_x_sale_order_id': id, 'default_x_sale_order_id': id}" invisible="not x_order_involves_room"/>
         </xpath>
         <xpath expr="//form[1]/sheet[1]/notebook[1]" position="inside">
-          <page string="Guests">
+          <page string="Guests" invisible="not x_order_involves_room">
             <field name="x_guest_line_ids">
               <list default_order="x_sequence asc,id desc" editable="bottom">
                 <field name="x_sequence" widget="handle"/>
-                <field optional="show" name="x_guest_partner_id" widget="res_partner_many2one" required="1"/>
+                <field optional="show" name="x_guest_partner_id" widget="many2one_avatar" required="1"/>
                 <field optional="show" name="x_guest_identity_check" widget="badge" readonly="True" decoration-success="x_guest_identity_check == 'ok'" decoration-warning="x_guest_identity_check == 'na'" decoration-danger="x_guest_identity_check == 'invalid'"/>
                 <field optional="show" name="x_room_resource_id" domain="[('default_role_id.x_is_a_room_offer', '!=', False), ('id', 'in', x_sol_resource_ids)]" options="{'no_create': True, 'no_create_edit': True}"/>
               </list>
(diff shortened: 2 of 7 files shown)


holiday_house/demo/sale_order_post.xml
```diff
@@ -47,12 +47,19 @@
         <field name="status">pickup</field>
     </record>
 
+    <record id="pickup_wizard_6" model="rental.order.wizard">
```
vava-odoo, line 50: why no version bump for this one? I wonder if a bump is needed to be honest, as only demo data are touched in industries. Only booking_engine manifest should be touched

booking_engine/data/ir_model_fields.xml
```diff
@@ -789,4 +789,40 @@ Note that saving a new survey with this set will archive previous survey with th
...
+        slot['x_booking_color'] = 1   # red - attention
+    else:
+        slot['x_booking_color'] = False # grey - no focus
+      ]]></field>
+  </record>
+  <record id="x_rental_status_field" model="ir.model.fields">
+    <field name="ttype">selection</field>
+    <field name="field_description">Rental Status</field>
+    <field name="model_id" ref="planning.model_planning_slot"/>
+    <field name="name">x_rental_status</field>
```
vava-odoo, line 823: I think the default_group_by works with a non-stored related. So I'd make this field readonly not-stored. Just need to adapt the domain where it is used

booking_engine/data/ir_ui_view.xml
```diff
@@ -1040,4 +1044,18 @@
...
   </record>
+  <record id="planning_search_view_inherit_booking_engine" model="ir.ui.view">
+    <field name="name">planning.slot.search.inherit.booking_engine</field>
+    <field name="model">planning.slot</field>
+    <field name="inherit_id" ref="planning.planning_view_search_base"/>
+    <field name="mode">extension</field>
+    <field name="active" eval="True"/>
+    <field name="arch" type="xml">
+        <xpath expr="//filter[@name='conflict_shifts']" position="before">
+          <filter string="Today's Check In" name="filter_checkin_today" domain="[('x_rental_status', '=', 'pickup'), ('sale_order_id.rental_start_date', '&gt;=', 'today'), ('sale_order_id.rental_start_date', '&lt;', 'today +1d')]"/>
```
vava-odoo: the spec says we should remove "my customer" filter. Did it drop from spec?

booking_engine/data/ir_ui_view.xml
```diff
@@ -649,7 +653,7 @@
         <field name="arch" type="xml">
             <xpath expr="/form/sheet/group" position="after">
                 <group>
-                    <group string="House Keeping" groups="booking_engine.group_house_keeping">
+                    <group string="House Keeping" groups="booking_engine.group_house_keeping" invisible="not role_ids">
```
vava-odoo: when I read the spec, I'd do this. What do you think?
```suggestion
                    <group string="House Keeping" groups="booking_engine.group_house_keeping" invisible="not x_is_stay_resource">
```

booking_engine/data/ir_model_fields.xml
```diff
@@ -789,4 +789,40 @@ Note that saving a new survey with this set will archive previous survey with th
...
+today_start = datetime.datetime.combine(datetime.date.today(), datetime.time.min)
+today_end = datetime.datetime.combine(datetime.date.today(), datetime.time.max)
+
+for slot in self:
+    order = slot.sale_order_id
+    if not order:
+        slot['x_booking_color'] = 5   # purple - internal
+        continue
+
+    if order.rental_status == 'pickup' and order.rental_start_date and today_start <= order.rental_start_date <= today_end:
```
vava-odoo:
```suggestion
    elif order.rental_status == 'pickup' and order.rental_start_date and today_start <= order.rental_start_date <= today_end:
```

booking_engine/data/ir_ui_view.xml
dhrs-odoo: No basically we want that when we create new resource we don't want to show house keeping section until we dont have role_ids related to a stay offer product

booking_engine/data/ir_ui_view.xml
dhrs-odoo: I didn't found that filter I discussed same with Savp

booking_engine/data/ir_ui_view.xml
vava-odoo: yes but role_ids is a standard field not related to stay offers at all

booking_engine/data/ir_ui_view.xml
```diff
@@ -254,9 +254,13 @@
   </record>
   <record id="planning_kanban_view" model="ir.ui.view">
     <field name="arch" type="xml">
-        <xpath expr="//field[@name='sale_line_id']" position="replace">
-          <field name="sale_order_id" display="full" widget="many2one" class="d-block" context="{'form_view_ref': 'booking_engine.rental_order_view', 'in_rental_app': 1}"/>
-        </xpath>
+      <xpath expr="//kanban" position="attributes">
+          <attribute name="default_group_by">x_rental_status</attribute>
```
vava-odoo, line 258: I'm not sure this should be done in the view itself, because then it impacts all planning view. I'd put that only for Booking app views

booking_engine/data/ir_model_fields.xml
```diff
@@ -789,4 +789,52 @@ Note that saving a new survey with this set will archive previous survey with th
...
+  <record id="x_has_room_offer_role_field" model="ir.model.fields">
+      <field name="name">x_has_room_offer_role</field>
+      <field name="field_description">Has Room Offer Role</field>
+      <field name="model_id" ref="resource.model_resource_resource"/>
+      <field name="ttype">boolean</field>
+      <field name="store" eval="False"/>
+      <field name="depends">role_ids.x_is_a_room_offer</field>
+      <field name="compute"><![CDATA[
+for record in self:
+    record['x_has_room_offer_role'] = any(role.x_is_a_room_offer for role in record.role_ids)
```
vava-odoo, line 837: To me, this is exactly the meaning of `x_is_stay_resource`, or I miss something.
Maybe @abdrahmanrashed could tell us who is wrong here 😄

dhrs-odoo: Yes, but if we move the computation to `x_is_stay_resource`, then we would need to make the field `store=False`. Because of that, a lot of additional changes would be required, as non-stored fields cannot be used in domains. So IMO, the current approach with this field looks fine to me.

what do you think ? @abdrahmanrashed

abdrahmanrashed: There is a new related field coming up that would fix problem, look at a similar server action example here: https://github.com/odoo/industry/pull/2061/changes#diff-93febc339301fc6802b7cc0388d2d72d64aeb6da6ee9c3788e17a37dc527e7beL134-L136

Since @Pierre-savp is on leave and it will take time for the PR to be tested, approved, and merged, I would suggest keep your change as is, but when it gets merged one of us should edit it in the other PR.

vava-odoo: so to my very basic question, which was: 
> are  `x_is_stay_resource` and `x_has_room_offer_role` equivalent?

you both answer no, is that correct?

dhrs-odoo: > so to my very basic question, which was:
> 
> > are  `x_is_stay_resource` and `x_has_room_offer_role` equivalent?
> 
> you both answer no, is that correct?

Yes

booking_engine/data/ir_model_fields.xml
```diff
@@ -789,4 +789,52 @@ Note that saving a new survey with this set will archive previous survey with th
...
+    elif order.rental_status == 'pickup' and order.rental_start_date and today_start <= order.rental_start_date <= today_end:
+        slot['x_booking_color'] = 3   # yellow - check in due
+    elif order.rental_status == 'return' and order.rental_return_date and order.rental_return_date > today_end:
+        slot['x_booking_color'] = 10  # green - ongoing
+    elif order.rental_status == 'return' and order.rental_return_date and today_start <= order.rental_return_date <= today_end:
+        slot['x_booking_color'] = 4   # blue - check out due
+    elif (order.rental_status == 'pickup' and order.is_late) or (order.rental_status == 'return' and order.is_late):
+        slot['x_booking_color'] = 1   # red - attention
+    else:
+        slot['x_booking_color'] = False # grey - no focus
```
abdrahmanrashed, line 816: nitpick, feel free to do whatever you want though...
```suggestion
    slot['x_booking_color'] = False   # grey - no focus
    if not order: slot['x_booking_color'] = 5   # purple - internal
    elif order.rental_status in ['pickup', 'return']:
        if order.is_late: slot['x_booking_color'] = 1   # red - attention
        elif order.rental_status == 'return':
            if order.rental_return_date > today_end:
            slot['x_booking_color'] = 10  # green - ongoing
            elif order.rental_return_date and today_start <= order.rental_return_date <= today_end:
            slot['x_booking_color'] = 4   # blue - check out due
        elif order.rental_start_date and today_start <= order.rental_start_date <= today_end:
            slot['x_booking_color'] = 3   # yellow - check in due
```

booking_engine/data/ir_model_fields.xml
```diff
@@ -789,4 +789,52 @@ Note that saving a new survey with this set will archive previous survey with th
...
+      ]]></field>
+  </record>
+  <record id="x_rental_status_field" model="ir.model.fields">
+    <field name="ttype">selection</field>
+    <field name="field_description">Rental Status</field>
+    <field name="model_id" ref="planning.model_planning_slot"/>
+    <field name="name">x_rental_status</field>
+    <field name="related">sale_order_id.rental_status</field>
+    <field name="store">False</field>
```
vava-odoo, line 825: does that really work? 😂

booking_engine/data/ir_model_fields.xml
vava-odoo: please no :upside_down_face:

vava-odoo (approved):

booking_engine/data/ir_model_fields.xml
vava-odoo: checked, yes :melting_face:

dhrs-odoo: yes it's working

vava-odoo:
@robodoo r+ squash

PR #2132, opened 2026-05-11
[IMP] headhunter: use window actions

vava-odoo:


vava-odoo:
@robodoo override=ci/industry
@robodoo r+

PR #2145, opened 2026-05-11
[FIX] test_generic: check modules installed in dependencies

abdrahmanrashed:
Before this commit the industry test would check if a module is installed or not iff it is included in its manifest. Now the tests also check the modules installed in other industry dependencies.


vava-odoo (approved):
@robodoo r+

vava-odoo:
@robodoo r+

PR #2149, opened 2026-05-11
[FIX] construction_developer: fix automation if no product on invoice lines

vava-odoo:
Before this commit, if an account move line was not linked to a product, the automation failed. This commit adds a check that ensures this action is only triggered for lines linked to a product.

opw-6201458


vava-odoo:
@robodoo r+

PR #2156, opened 2026-05-11
[FIX] beverage_distributor, micro_brewery: Automated BoM

jowes-odoo:
Currently, the automation rule "Automated BoM" created with beverage_distributor and micro_brewery causes an error in a multi-company database.

The causing lines in the automaton rule are:
```
'route_id': env.ref('mrp.route_warehouse0_manufacture').id,
'location_id': env.ref('stock.stock_location_stock').id,
```
These two external identifiers are created for the first company in the database. If this automation were to run while not in the first database, or for products specifically not in company 1, this will cause an error.

The solution is to dynamically find the route and location depending on the current company.

Steps to reproduce:

1. Install beverage_distributor or micro_brewery
2. Create and select a new company
3. Create Product A
4. Create Product B
5. Set Product B to contain 10 Product As
6. Save > Error `Incompatible companies on records`

opw-6095558
closes #2156


jowes-odoo:
@robodoo up to saas-18.4

PR #2163, opened 2026-05-12
[FIX] headhunter: fix demo attachment and CV

vava-odoo:
Before this commit, demo attachments were not properly linked to hr.applicant records, leading to traceback when browsing the ids (falsy values are prohibited since https://github.com/odoo/odoo/commit/4290724a4c8c57fba4f4d3d688d38f65dadcc38f).

This commit properly links ir.attachment and hr.applicant to avoid any issues with demo data.

task-6208544


headhunter/demo/website_ir_attachment.xml
dhrs-odoo, line 20: `res_id` should be added over here no?

dhrs-odoo:
Hello @vava-odoo 
LGTM :rocket: 
Could you please add a proper commit message so we have a clear understanding in the future of why these changes were made? Also, it should be a task instead of an OPW

vava-odoo:
@dhrs-odoo thanks for the review. Applied suggestions (and removed unused old ir.attachment). Could you check again?

dhrs-odoo:
Hello @vava-odoo 
Thanks :innocent: 
Good to go :rocket:

vava-odoo:
@robodoo r+

PR #2165, opened 2026-05-12
[IMP] sport_events: remove sport events industry

mega-odoo:
The sports events industry has been removed, as users in this domain typically look for an event management software, but not a sport event management software.

Taks Id - 6153269


vava-odoo:
@robodoo override=ci/industry
@robodoo r+

PR #2168, opened 2026-05-13
[FIX] *: prevent an error while opening POS session

dhgoy-odoo:
*= ['bowling', 'food_trucks', 'micro_brewery', 'sports_club']

Currently, an error occurs while opening the restaurant session.

Step to reproduce:
- Install the bowling module.
- Open the Restaurant POS session.

`TypeError: 'str' object is not a mapping`

The issue occurs because floor_plan_layout values are defined as strings in XML
data. During floor plan computation, the value is later unpacked as a
dictionary [1], which causes the traceback 

link [1]: https://github.com/odoo/odoo/blob/76e179f24807612f2d864f14f1a0a1a5661585b0/addons/pos_restaurant/models/pos_config.py#L186

To fix this issue, define floor_plan_layout values using eval="{}" in the XML
data, so they are stored as dictionaries instead of strings.

The same issue also exists in food_trucks, micro_brewery, and sports_club, but
no traceback occurs because module_pos_restaurant(Is a Bar/Restaurant) is
disabled (False), which skips the restaurant floor plan computation. If
that is enabled (True), the same error will occur.

Task ID-6193639

### DIFF
--- a/bowling/__manifest__.py
+++ b/bowling/__manifest__.py
@@ -1,5 +1,6 @@
 {
     'name': 'Bowling Alley',
+    'version': '1.1',
     'category': 'Hospitality',
     'depends': [
         'base_industry_data',
--- a/bowling/data/restaurant_table.xml
+++ b/bowling/data/restaurant_table.xml
@@ -3,133 +3,133 @@
   <record id="restaurant_table_1" model="restaurant.table">
     <field name="floor_id" ref="restaurant_floor_1"/>
     <field name="table_number">1</field>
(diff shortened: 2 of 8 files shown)


dhrs-odoo (approved):
Hello @dhgoy-odoo 
Great Work 👏 
Just one comment to bump the minor version and we are good to go 🚀

vava-odoo (approved):
@robodoo r+
Well done @dhgoy-odoo !

PR #2170, opened 2026-05-13
[IMP] headhunter: configurable mail templates

dhrs-odoo:
Before this PR, the CV sending flow relied on a static mail template. As a result, users could not use a different template without modifying the server action manually.

With this commit, a boolean field has been added on mail templates to allow users to define which template should be used for sending CVs, making the flow more flexible and configurable.

**Task - 6213718**

### DIFF
--- a/headhunter/data/ir_actions_act_window.xml
+++ b/headhunter/data/ir_actions_act_window.xml
@@ -65,4 +65,12 @@
         <field name="view_mode">list,form</field>
         <field name="context">{'group_by': 'job_id'}</field>
     </record>
+    <record id="email_template_window_action" model="ir.actions.act_window">
+        <field name="name">Email Templates</field>
+        <field name="res_model">mail.template</field>
+        <field name="view_mode">form,list</field>
+        <field name="view_id" ref="mail.email_template_tree" />
+        <field name="search_view_id" ref="mail.view_email_template_search"/>
+        <field name="context">{'search_default_filter_applicant': 1}</field>
+    </record>
 </odoo>
--- a/headhunter/data/ir_model_fields.xml
+++ b/headhunter/data/ir_model_fields.xml
@@ -176,6 +176,13 @@ for job in self:
         <field name="ttype">boolean</field>
         <field name="copied" eval="False"/>
     </record>
+    <record id="x_is_template_to_send_cv_field" model="ir.model.fields">
+        <field name="name">x_is_template_to_send_cv</field>
+        <field name="field_description">Use for Send CV button</field>
+        <field name="model_id" ref="mail.model_mail_template" />
+        <field name="ttype">boolean</field>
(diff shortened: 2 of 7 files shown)


headhunter/data/ir_model_fields.xml
```diff
@@ -176,6 +176,13 @@ for job in self:
         <field name="ttype">boolean</field>
         <field name="copied" eval="False"/>
     </record>
+    <record id="x_use_for_cv_sent_field" model="ir.model.fields">
+        <field name="name">x_use_for_cv_sent</field>
```
vava-odoo: Could you use a better formulation here? something like `x_is_template_to_send_cv`

headhunter/data/ir_model_fields.xml
```diff
@@ -176,6 +176,13 @@ for job in self:
...
+        <field name="field_description">Use for Send CV button</field>
+        <field name="model_id" ref="mail.model_mail_template" />
+        <field name="ttype">boolean</field>
+        <field name="help">If this toggle is enabled on multiple email templates, the most recently updated template will be used for the 'Send CV' button.</field>
```
vava-odoo: The default order for `mail.template` is `'user_id, name, id'`. So "_the most recently updated template_" is not correct.

headhunter/data/ir_ui_view.xml
```diff
@@ -281,4 +281,26 @@
...
         </field>
     </record>
+    <record id="view_email_template_search_inherit" model="ir.ui.view">
+        <field name="name">mail.template.search.inherit.view</field>
+        <field name="model">mail.template</field>
+        <field name="inherit_id" ref="mail.view_email_template_search"/>
+        <field name="active" eval="True"/>
+        <field name="arch" type="xml">
+            <xpath expr="//filter[@name='base_templates']" position="after">
+                <filter string="Applicant Templates" name="filter_applicant" domain="[('model', '=', 'hr.applicant')]"/>
```
vava-odoo, line 291: This filter is supposed to be used by the new "email template" menu, but I don't think you added it.

vava-odoo (changes requested):

headhunter/data/ir_actions_act_window.xml
```diff
@@ -65,4 +65,7 @@
         <field name="view_mode">list,form</field>
         <field name="context">{'group_by': 'job_id'}</field>
     </record>
+    <record id="mail.action_email_template_tree_all" model="ir.actions.act_window" forcecreate="0">
```
vava-odoo: wouldn't it make more sense to create a new window action here?

vava-odoo (approved):

headhunter/data/ir_actions_act_window.xml
dhrs-odoo: No everything would be same just like base window action for that reason I haven't made new window action 
EDIT -  yes will add new window action otherwise we need to add `uninstall_hook_base_automation` and `server_action`

vava-odoo:
@robodoo r+

PR #2171, opened 2026-05-13
[IMP] booking_engine: add Daily Lines on SO and Manager Dashboard

frva-odoo:
This PR adds a dashboard to visualize informations for an hotel
such as occupancy rate or average daily rate. This dashboard uses a new
model linked to sale orders, x_so_daily_line, that are represented one
for each room and for each day of stayover.

Task-6176659

### DIFF
--- a/booking_engine/__manifest__.py
+++ b/booking_engine/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Booking',
-    'version': '1.20',
+    'version': '1.21',
     'category': 'Hospitality',
     'author': 'Odoo S.A.',
     'depends': [
--- a/booking_engine/__manifest__.py
+++ b/booking_engine/__manifest__.py
@@ -44,6 +44,7 @@
         'data/survey_survey.xml',
         'data/survey_question.xml',
         'data/survey_question_answer.xml',
+        'data/spreadsheet_dashboard.xml',
     ],
     'demo': [
         'demo/website_menu.xml',
--- a/booking_engine/data/base_automation.xml
+++ b/booking_engine/data/base_automation.xml
@@ -158,4 +158,53 @@
         <field name="trigger_field_ids" eval="[(6, 0, [ref('x_is_customer_satisfaction_survey_field')])]"/>
         <field name="action_server_ids" eval="[(6, 0, [ref('server_action_to_archive_surveys')])]"/>
     </record>
+
+    <!-- Daily Lines -->
+    <record id="automation_daily_lines_on_so_create" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_create_daily_lines_from_so')])]"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="filter_domain">[('rental_status', '=', 'pickup'), ('rental_start_date', '!=', False), ('rental_return_date', '!=', False)]</field>
+        <field name="filter_pre_domain">[('rental_status', 'in', ['draft', 'sent'])]</field>
+        <field name="name">Daily Lines: On SO confirmation</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('sale_renting.field_sale_order__rental_status')])]"/>
+    </record>
+    <record id="automation_daily_lines_on_so_cancel" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_unlink_daily_lines_from_so')])]"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="filter_domain">[('x_daily_lines_ids', '!=', False), ('state', '=', 'cancel')]</field>
+        <field name="name">Daily Lines: On SO cancelled</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('sale.field_sale_order__state')])]"/>
+    </record>
+    <record id="automation_daily_lines_on_so_delete" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_unlink_daily_lines_from_so_2')])]"/>
+        <field name="trigger">on_unlink</field>
+        <field name="filter_domain">[('x_daily_lines_ids', '!=', False)]</field>
+        <field name="name">Daily Lines: On SO delete</field>
+    </record>
+    <record id="automation_daily_lines_on_sol_create" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_create_daily_lines_from_sol')])]"/>
+        <field name="trigger">on_create</field>
+        <field name="filter_domain">[('product_id.product_tmpl_id.x_is_a_room_offer', '=', True), ('order_id.rental_status', 'in', ['pickup', 'return', 'returned']), ('order_id.rental_start_date', '!=', False), ('order_id.rental_return_date', '!=', False)]</field>
+        <field name="name">Daily Lines: On SOL creation in confirmed order</field>
+    </record>
+    <record id="automation_daily_lines_on_sol_write" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_modify_daily_lines_from_sol')])]"/>
+        <field name="trigger">on_write</field>
+        <field name="filter_domain">[('x_daily_line_ids', '!=', False)]</field>
+        <field name="name">Daily Lines: On SOL with daily lines write</field>
+    </record>
+    <record id="automation_daily_lines_on_so_rental_date_changed" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_modify_daily_lines_from_so_date_change')])]"/>
+        <field name="trigger">on_write</field>
+        <field name="filter_domain">[('rental_status', 'in', ['pickup', 'return']), ('rental_start_date', '!=', False), ('rental_return_date', '!=', False)]</field>
+        <field name="filter_pre_domain">[('rental_status', 'in', ['pickup', 'return'])]</field>
+        <field name="name">Daily Lines: On SO rental date changed</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('sale_renting.field_sale_order__rental_start_date'), ref('sale_renting.field_sale_order__rental_return_date')])]"/>
+    </record>
 </odoo>
--- a/booking_engine/data/ir_actions_server.xml
+++ b/booking_engine/data/ir_actions_server.xml
@@ -528,4 +528,104 @@ if survey:
             invite.action_invite()
         orders_to_survey.write({'x_is_survey_sent': True})]]></field>
     </record>
+
+    <!-- Daily Lines -->
+    <record id="server_action_create_daily_lines_from_so" model="ir.actions.server">
+        <field name="code"><![CDATA[
+dates_covered = []
+current_date = record.rental_start_date
+while current_date < record.rental_return_date:
+    dates_covered.append(current_date.date())
+    current_date += datetime.timedelta(days=1)
+env.ref('booking_engine.server_action_method_create_daily_lines').with_context(
+    active_model='sale.order.line', active_ids=records.order_line.filtered(lambda l: l.product_id.product_tmpl_id.x_is_a_room_offer).ids, dates=dates_covered
+).run()
+]]></field>
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Dashboard: Create daily lines from so</field>
+    </record>
+    <record id="server_action_unlink_daily_lines_from_so" model="ir.actions.server">
+        <field name="code"><![CDATA[
+records.x_daily_lines_ids.unlink()
+]]></field>
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Dashboard: Delete daily lines from so</field>
+    </record>
+    <record id="server_action_unlink_daily_lines_from_so_2" model="ir.actions.server">
+        <field name="code"><![CDATA[
+records.x_daily_lines_ids.unlink()
+]]></field>
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Dashboard: Delete daily lines from so 2</field>
+    </record>
+    <record id="server_action_create_daily_lines_from_sol" model="ir.actions.server">
+        <field name="code"><![CDATA[
+dates_covered = []
+current_date = record.order_id.rental_start_date
+while current_date < record.order_id.rental_return_date:
+    dates_covered.append(current_date.date())
+    current_date += datetime.timedelta(days=1)
+env.ref('booking_engine.server_action_method_create_daily_lines').with_context(
+    active_model='sale.order.line', active_ids=records.ids, dates=dates_covered
+).run()
+]]></field>
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="state">code</field>
+        <field name="name">Dashboard: Create daily lines from sol</field>
+    </record>
+    <record id="server_action_method_create_daily_lines" model="ir.actions.server">
+        <field name="code"><![CDATA[
+dates_covered = env.context.get('dates') or []
+daily_lines_values = []
+for record in (records or []):
+    for current_date in dates_covered:
+        daily_lines_values.append({
+            'x_sale_order_id': record.order_id.id,
+            'x_sale_order_line_id': record.id,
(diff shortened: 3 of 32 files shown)


hotel/data/spreadsheet_dashboard.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<odoo noupdate="1">
+    <record id="booking_engine.spreadsheet_dashboard_group_accomodation" model="spreadsheet.dashboard.group" forcecreate="0">
+        <field name="name">Hotel</field>
```
vava-odoo, line 4: This will only work in English. All translations will take the booking_engine term translation...

hotel/data/knowledge_article.xml
```diff
@@ -418,6 +418,17 @@
             <li>The Gantt chart is color-coded to reflect on the availability for you to work on filling in.</li>
             <li>Access to more details clicking the slot.</li>
         </ul>
+        <h4>Monitoring performance</h4>
+        <p>Because keeping an eye on your business performance matter, the module comes with a dedicated dashboard.</p>
```
vava-odoo:
```suggestion
        <p>Because keeping an eye on your business performance matters, the module comes with a dedicated dashboard.</p>
```

booking_engine/data/ir_model_access.xml
```diff
@@ -27,4 +27,13 @@
         <field name="perm_write" eval="False"/>
         <field name="perm_unlink" eval="False"/>
     </record>
+    <record id="x_model_so_daily_line_access" model="ir.model.access">
```
vava-odoo: Do we want anyone to be able to modify those values 🤔

booking_engine/data/ir_model_fields.xml
```diff
@@ -785,4 +785,98 @@ Note that saving a new survey with this set will archive previous survey with th
...
+    <record id="_field_daily_line" model="ir.model.fields">
+        <field name="ttype">integer</field>
+        <field name="field_description">Sequence</field>
+        <field name="model_id" ref="x_model_so_daily_line"/>
+        <field name="name">x_sequence</field>
+    </record>
+    <record id="x_sale_order_id_field_daily_line" model="ir.model.fields">
+        <field name="ttype">many2one</field>
+        <field name="copied" eval="True"/>
+        <field name="field_description">X Sale Order</field>
```
vava-odoo:
```suggestion
        <field name="field_description">Sale Order</field>
```

booking_engine/data/ir_model_fields.xml
```diff
@@ -785,4 +785,98 @@ Note that saving a new survey with this set will archive previous survey with th
       <field name="model_id" ref="sale.model_sale_order"/>
       <field name="ttype">boolean</field>
   </record>
+
+    <!-- Daily Lines -->
+    <record id="_field_daily_line" model="ir.model.fields">
+        <field name="ttype">integer</field>
+        <field name="field_description">Sequence</field>
+        <field name="model_id" ref="x_model_so_daily_line"/>
+        <field name="name">x_sequence</field>
```
vava-odoo: Is this field useful? Isn't it ordered by date anyway?

booking_engine/data/ir_model_fields.xml
```diff
@@ -785,4 +785,98 @@ Note that saving a new survey with this set will archive previous survey with th
...
+    </record>
+    <record id="x_sale_order_id_field_daily_line" model="ir.model.fields">
+        <field name="ttype">many2one</field>
+        <field name="copied" eval="True"/>
```
vava-odoo, line 840: do we really want to copy  all this fields?

booking_engine/data/ir_model_fields.xml
```diff
@@ -785,4 +785,98 @@ Note that saving a new survey with this set will archive previous survey with th
...
+        <field name="name">x_price</field>
+        <field name="currency_field">x_currency_id</field>
+    </record>
+    <record id="x_state_field_daily_line" model="ir.model.fields">
+        <field name="ttype">selection</field>
+        <field name="copied" eval="True"/>
+        <field name="field_description">State</field>
+        <field name="model_id" ref="x_model_so_daily_line"/>
+        <field name="name">x_state</field>
+        <field name="selection">[('planned', 'Planned'), ('done', 'Done'), ('cancel', 'Cancel')]</field>
```
vava-odoo: need to be defined in ir.model.fields.selection records

booking_engine/data/base_automation.xml
```diff
@@ -174,4 +174,53 @@
         <field name="trigger_field_ids" eval="[(6, 0, [ref('x_is_customer_satisfaction_survey_field')])]"/>
         <field name="action_server_ids" eval="[(6, 0, [ref('server_action_to_archive_surveys')])]"/>
     </record>
+
+    <!-- Daily Lines -->
+    <record id="automation_daily_lines_on_so_create" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_create_daily_lines_from_so')])]"/>
+        <field name="trigger">on_create_or_write</field>
+        <field name="filter_domain">[('rental_status', '=', 'pickup'), ('rental_start_date', '!=', False), ('rental_return_date', '!=', False)]</field>
```
vava-odoo, line 167: couldn't the order move to 'return' (picked-up) status directly?

booking_engine/data/ir_actions_server.xml
```diff
@@ -561,4 +561,100 @@ if survey:
             invite.action_invite()
         orders_to_survey.write({'x_is_survey_sent': True})]]></field>
     </record>
+
+    <!-- Daily Lines -->
+    <record id="server_action_create_daily_lines_from_so" model="ir.actions.server">
+        <field name="code"><![CDATA[
+for line in record.order_line:
```
vava-odoo: maybe need to loop on records? (if we confirm two RO)

booking_engine/data/ir_actions_server.xml
```diff
@@ -561,4 +561,100 @@ if survey:
...
+        <field name="code"><![CDATA[
+records.x_daily_lines_ids.unlink()
+]]></field>
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Dashboard: Delete daily lines from so when cancelled</field>
+    </record>
+    <record id="server_action_unlink_daily_lines_from_so" model="ir.actions.server">
+        <field name="code"><![CDATA[
+records.x_daily_lines_ids.unlink()
```
vava-odoo, line 558: Same server action

booking_engine/data/ir_actions_server.xml
```diff
@@ -561,4 +561,100 @@ if survey:
...
+]]></field>
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Dashboard: Delete daily lines from so</field>
+    </record>
+    <record id="server_action_create_daily_lines_from_sol" model="ir.actions.server">
+        <field name="code"><![CDATA[
+current_date = record.order_id.rental_start_date
```
vava-odoo, line 567: same: couldn't we create 2 SOL at a time?

booking_engine/data/ir_actions_server.xml
```diff
@@ -561,4 +561,100 @@ if survey:
...
+        while current_date < record.rental_return_date:
+            price_for_day = record.pricelist_id._get_product_price(line.product_id, 1.0, date=current_date)
+            env['x_so_daily_line'].create({
+                'x_sale_order_id': record.id,
+                'x_sale_order_line_id': line.id,
+                'x_date': current_date,
+                'x_price': price_for_day,
+                'x_state': 'planned',
+            })
+            current_date += datetime.timedelta(days=1)
```
vava-odoo: this part is duplicated in another server action. Couldn't we replace this part by calling the other server action?

booking_engine/data/base_automation.xml
```diff
@@ -174,4 +174,53 @@
...
+    </record>
+    <record id="automation_daily_lines_on_so_rental_date_changed" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_modify_daily_lines_from_so_date_change')])]"/>
+        <field name="trigger">on_write</field>
+        <field name="filter_domain">[('rental_status', 'in', ['pickup', 'return']), ('rental_start_date', '!=', False), ('rental_return_date', '!=', False)]</field>
+        <field name="filter_pre_domain">[('rental_status', 'in', ['pickup', 'return'])]</field>
+        <field name="name">Daily Lines: On SO rental date changed</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('sale_renting.field_sale_order__rental_start_date'), ref('sale_renting.field_sale_order__rental_return_date')])]"/>
+    </record>
```
vava-odoo, line 209: no need for SOL unlink?

booking_engine/data/ir_actions_server.xml
```diff
@@ -561,4 +561,100 @@ if survey:
...
+    represented_dates = record.x_daily_lines_ids.mapped('x_date')
+    to_cover_dates = []
+    current_date = record.rental_start_date
+    while current_date < record.rental_return_date:
+        if current_date.date() not in represented_dates:
+            to_cover_dates.append(current_date.date())
+        current_date += datetime.timedelta(days=1)
+    for line in record.order_line.filtered(lambda l: l.product_id.product_tmpl_id.x_is_a_room_offer):
+        for date in to_cover_dates:
+            price_for_day = record.pricelist_id._get_product_price(line.product_id, 1.0, date=date)
```
vava-odoo: same comment about duplicated code

vava-odoo (commented):
Here are a few comments.
My major point is: is that the good strategy? Shouldn't it be computed on the fly at dashboard opening? Shouldn't it be done in sql?
Other aspect: do we want this to be added to all 4 industries? It seems interesting only for hotel

booking_engine/data/base_automation.xml
frva-odoo: We don't need it as it is forbidden to erase a sol on a confirmed SO, and daily lines are not generated before confirmation

booking_engine/data/base_automation.xml
frva-odoo: We only want to generate in the case they were in these states before, not if it was from a later and came back so I think this is the right way to go. What do you mean ?

booking_engine/data/ir_model_fields.xml
frva-odoo: Seems like the most important to copy to me; if I duplicate a daily line, it's often to keep the same so linked

hotel/data/spreadsheet_dashboard.xml
frva-odoo: Whyyyy ???? :cry:

booking_engine/data/base_automation.xml
vava-odoo: I mean: could it go from confirm to return (without being in pickup state). Real question

hotel/data/spreadsheet_dashboard.xml
vava-odoo: because that is how translated terms work... the translation will look into the po(t) file(s) from the original module only.

booking_engine/data/ir_actions_server.xml
```diff
@@ -560,4 +561,104 @@ if survey:
...
+    </record>
+    <record id="server_action_unlink_daily_lines_from_so" model="ir.actions.server">
+        <field name="code"><![CDATA[
+records.x_daily_lines_ids.unlink()
+]]></field>
+        <field name="model_id" ref="sale.model_sale_order"/>
+        <field name="state">code</field>
+        <field name="name">Dashboard: Delete daily lines from so</field>
+    </record>
+    <record id="server_action_unlink_daily_lines_from_so_2" model="ir.actions.server">
```
vava-odoo, line 556: so you got into same trouble as Tristan

booking_engine/data/base_automation.xml
```diff
@@ -174,4 +174,53 @@
...
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_create_daily_lines_from_sol')])]"/>
+        <field name="trigger">on_create</field>
+        <field name="filter_domain">[('product_id.product_tmpl_id.x_is_a_room_offer', '=', True), ('order_id.rental_status', 'in', ['pickup', 'return', 'returned']), ('order_id.rental_start_date', '!=', False), ('order_id.rental_return_date', '!=', False)]</field>
+        <field name="name">Daily Lines: On SOL creation in confirmed order</field>
+    </record>
+    <record id="automation_daily_lines_on_sol_write" model="base.automation">
+        <field name="model_id" ref="sale.model_sale_order_line"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('server_action_modify_daily_lines_from_sol')])]"/>
+        <field name="trigger">on_write</field>
+        <field name="filter_domain">[('x_daily_line_ids', '!=', False)]</field>
```
vava-odoo, line 198: wouldn't this be costly? Triggered at any write on SOL

vava-odoo (approved):

booking_engine/data/ir_actions_server.xml
frva-odoo: Thanks to him, I wasn't surprised by the issue

booking_engine/data/base_automation.xml
frva-odoo: Yes But I don't see any other way to do it efficiently, and this is an issue for sale order with a lot of daily lines, which happens for:
- a lot of bookings of different room in one SO
- a really really long stay
I think that is not really common and the cost of the server action is not that big for it to be an issue

frva-odoo:
> Here are a few comments. My major point is: is that the good strategy? Shouldn't it be computed on the fly at dashboard opening? Shouldn't it be done in sql? Other aspect: do we want this to be added to all 4 industries? It seems interesting only for hotel

The reason why we cannot do that is because we want the price for each day at the moment of the SO, so the product price cannot be used and the sale order line price either as, since 19.4, the price may vary day to day by the price rules so sol price =  average_price_during_period. So we need a custom model to remember the prices for each day.

Considering the usage, as it may not be relevant for everyone, we should add settings that will be done at the same time as availabilities settings from the task 6289886, and that will be done in this PR so we can merge the setting at the same time as the feature

hotel/data/spreadsheet_dashboard.xml
frva-odoo: Then is it not possible to change the name of the dashboard from one app to the other ?

hotel/data/res_config_settings.xml
```diff
@@ -13,4 +13,12 @@
             </xpath>
         </field>
     </record>
+    <data noupdate="1">
+        <record model="res.config.settings" id="res_config_settings_enable">
+            <field name="x_module_house_keeping" eval="1"/>
+            <field name="x_setting_approvers" eval="1"/>
```
Pierre-savp: this one we don't want

hotel/data/res_config_settings.xml
```diff
@@ -13,4 +13,11 @@
             </xpath>
         </field>
     </record>
+    <data noupdate="1">
+        <record model="res.config.settings" id="res_config_settings_enable">
+            <field name="x_module_house_keeping" eval="1"/>
```
vava-odoo, line 18: it seems to be on by default already 🤔

booking_engine/__manifest__.py
```diff
@@ -44,6 +44,7 @@
         'data/survey_survey.xml',
         'data/survey_question.xml',
         'data/survey_question_answer.xml',
+        'data/spreadsheet_dashboard.xml',
```
vava-odoo, line 47: missing bump

booking_engine/data/spreadsheet_dashboard.xml
```diff
@@ -0,0 +1,13 @@
...
+<odoo noupdate="1">
+    <record id="spreadsheet_dashboard_group_accomodation" model="spreadsheet.dashboard.group">
+        <field name="name">Accomodation</field>
+        <field name="sequence">0</field>
+    </record>
+    <record id="spreadsheet_dashboard_1" model="spreadsheet.dashboard">
+        <field name="spreadsheet_snapshot" type="base64" file="booking_engine/static/src/binary/spreadsheet_dashboard/PerformanceDashboard.osheet.json"/>
+        <field name="name">Performance</field>
+        <field name="dashboard_group_id" ref="spreadsheet_dashboard_group_accomodation"/>
+        <field name="group_ids" eval="[(6, 0, [ref('base.group_system')])]"/>
```
vava-odoo, line 11: why group_system 🤔 do we assume managers = admin?

booking_engine/data/res_config_settings.xml
vava-odoo: This is probably something you'd want to edit, no?

vava-odoo (approved):

booking_engine/data/spreadsheet_dashboard.xml
frva-odoo: This is what SAVP defined in 6289886, I guess for now to keep it simple we don't make a custom group for this, and let only the admin see this

vava-odoo:
@robodoo rebase-ff

vava-odoo:
@robodoo r+

vava-odoo:
@robodoo r+

vava-odoo:
@robodoo r-

PR #2173, opened 2026-05-14
[19.0] [FIX] construction_developer: correctly duplicate work items on SO

### LINKED ISSUE #2172: [19.0] construction_developer: Duplicating sale.order updates original order pricing via work item
**Impacted versions:**
- 19.0 (Addon: `construction_developer`)

**Steps to reproduce:**
1. Install the `construction_developer` application.
2. Create a new Sales Order (`sale.order`) and add a work item via a template product.
3. Duplicate the Sales Order.
4. Open the duplicated Sales Order and edit its work item.

**Current behavior:**
Editing the work item on the duplicated Sales Order alters the pricing and lines of the original Sales Order. 

**Technical Analysis & Root Cause:**
The issue stems from how the Work Item relationship is defined on `sale.order.line` and Odoo's default `copy()` behavior.
1. The field `x_line_work_item_id` on `sale.order.line` is defined in `construction_developer/data/ir_model_fields.xml` as a `Many2one` relation to `x_work_item`.
2. By default, Odoo sets `copy=True` for `Many2one` fields.
3. When `sale.order.copy()` is called, it recursively calls `copy()` on its `sale.order.line` dependencies. The reference `x_line_work_item_id` is copied verbatim.
4. Both the old `sale.order.line` and the new duplicated `sale.order.line` now reference the **exact same** `x_work_item` record.
5. Furthermore, the `x_work_item` record itself contains a tracked `x_sale_order_line_id` reference back to the original `sale.order.line`. 
6. When a user modifies the shared `x_work_item` (e.g., changes its unit cost/price), the automated actions (`base_automation_work_item_on_write_unit_cost_price_change`, etc.) trigger updates that get pushed to the recorded `x_sale_order_line_id` (the original line).

**Proposed Solution:**
This can be resolved in two ways depending on the expected business logic:
- **Approach 1 (Studio/XML approach - Recreate from Template):** Set `copied=False` for the `x_line_work_item_id` field in `ir_model_fields.xml`. 
  ```xml
  <record id="x_line_work_item_id_field_sale_order_line" model="ir.model.fields">
      ...
      <field name="copied" eval="False"/>
  </record>
  ```
  When the line is duplicated, `x_line_work_item_id` will be `False`. This successfully triggers the existing `base_automation_sol_on_write` automation, which detects the missing work item and creates a fresh instance from the product's `x_work_item_template_id`. Note: this causes the duplicate quote to lose any manual customizations done to the original quote's work item.
- **Approach 2 (Python approach - Deep Duplication):** Introduce a python override for `sale.order.line`'s `copy_data()` or `copy()` method in a new `models/` directory, specifically ensuring that when `sale.order.line` is copied, it performs a deep `copy()` on its referenced `x_line_work_item_id` rather than a standard reference assignment. This preserves all manual modifications of the customized work item across the duplicated quote.

vvro:
Fixes #2172.

When a sale.order is duplicated, the `Many2one` references are copied natively, meaning both the old and new sale.order.line referenced the exact same `x_work_item`. Modifying the work item on the new duplicate altered the pricing on the original SO.

This uses the Odoo Online compatible XML approach. By adding `copied=False` to the field definition in Studio (`ir_model_fields.xml`), the work item reference is appropriately stripped upon duplication. This subsequently triggers the existing standard automation `base_automation_sol_on_write` to re-instantiate a fresh Work Item directly from the product template.

### DIFF
--- a/construction_developer/data/ir_model_fields.xml
+++ b/construction_developer/data/ir_model_fields.xml
@@ -367,6 +367,7 @@ for record in self:
         <field name="model_id" ref="sale.model_sale_order_line"/>
         <field name="name">x_line_work_item_id</field>
         <field name="relation">x_work_item</field>
+        <field name="copied" eval="False"/>
     </record>
 
     <record id="x_work_item_template_id_field_product_product" model="ir.model.fields">


vava-odoo (approved):
Indeed, many thanks for the fix!
@robodoo r+

PR #2176, opened 2026-05-15
[FIX] booking_engine: active filter group_by instead of context value

vava-odoo:
Steps to reproduce:
- Go to booking engine app > orders > open an order
- Open the product catalog
- Traceback

Reason:
the 'group_by' was kept in the context, although the view was changed to another model (product.template instead of sale.order). This leads to an error because the field (rental_status) does not exist on this model.

This commit fixes the issue by activating a group_by filter on the Orders view by default instead of hardcoding the groupby key in the context.

task-6218176

### DIFF
--- a/booking_engine/__manifest__.py
+++ b/booking_engine/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Booking Engine',
-    'version': '1.13',
+    'version': '1.14',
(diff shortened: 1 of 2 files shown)


dhrs-odoo (approved):
Hello @vava-odoo 
LGTM 🚀

vava-odoo:
@robodoo r+

PR #2185, opened 2026-05-19
[FIX] *: prevent an error while validating the payment

dhgoy-odoo:
*= ['florist', 'bike_shop', 'museum', 'takeaway_restaurant', 'wine_merchant']

Currently, an error occurs while validating a POS payment in above industries.

Step to reproduce:

- Install the florist industry.
- Open a POS session.
- Validate a payment.

ValidationError: Journal is required for payment methods.

The issue occurs because some POS payment methods are created without a
journal_id. The journal_id field is mandatory for payment methods,
and payment validation expects valid journals on all configured payment
methods.

To fix this issue, create a cash journal and assign it to the cash payment
method, and assign a bank journal to the card payment method.

In the takeaway restaurant, the same cash payment method was used across
multiple POS configurations. If a journal is assigned to the shared cash payment
method, it raises a validation error because the same cash payment method
cannot be used in different POS configurations. To fix this issue, create
separate cash journals for different POS configurations and assign them to
their respective cash payment methods.

Task ID-6193655

### DIFF
--- a/bike_shop/__manifest__.py
+++ b/bike_shop/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Bike Shop',
-    'version': '1.1',
+    'version': '1.2',
     'category': 'Retail',
     'author': 'Odoo S.A.',
     'depends': [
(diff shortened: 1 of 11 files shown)


vava-odoo:
solved elsewhere

PR #2187, opened 2026-05-19
[FIX] booking_engine: fix day added on rental order confirm

frva-odoo:
This commit fixes the incorrect behavior where confirming a rental order corrects the time period of the order incorrectly. It should only change the hours to correspond to the arrival and departure hours defined, but this also currently set the departure date to 1 day later than what we defined.

This is now fixed by taking timezones into account, so that the
changes are not modifying the days unwillingly.

TASK-6218267


frva-odoo:
From SAVP: I should remove the days offset, the hours don't matter when you define a start_time/end_time for a rental so

vava-odoo:
task cancelled, server actions removed

PR #2189, opened 2026-05-19
[FIX] booking_engine: prevent traceback while saving rental order

dhrs-odoo:
Before this commit, saving a rental order after creating shifts raised a traceback, preventing the order from being saved.

**Cause :**
- _compute_name() was explicitly called from the server action using record.order_line._compute_name(), even though it is already triggered internally by Odoo.

**Fix :**
- We removed the line as it was already been called priorly and calling it again makes  no sense

**OPW - 6218502**

### DIFF
--- a/booking_engine/data/ir_actions_server.xml
+++ b/booking_engine/data/ir_actions_server.xml
@@ -60,7 +60,6 @@ for record in records:
     if pickup_time: record['rental_start_date'] = record.rental_start_date.replace(hour=int(pickup_time), minute=int(pickup_time % 1 * 60))
     if return_time: record['rental_return_date'] = record.rental_return_date.replace(hour=int(return_time), minute=int(return_time % 1 * 60))
     record.action_update_rental_prices()
-    record.order_line._compute_name()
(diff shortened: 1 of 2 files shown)


booking_engine/data/ir_actions_server.xml
```diff
@@ -60,7 +60,6 @@ for record in records:
     if pickup_time: record['rental_start_date'] = record.rental_start_date.replace(hour=int(pickup_time), minute=int(pickup_time % 1 * 60))
     if return_time: record['rental_return_date'] = record.rental_return_date.replace(hour=int(return_time), minute=int(return_time % 1 * 60))
     record.action_update_rental_prices()
-    record.order_line._compute_name()
```
abdrahmanrashed, line 63: Hello @dhrs-odoo, you are certainly correct this function is being called internally after the `action_update_rental_prices` call. This wasn't the case back when it was added, that is why it is there.

abdrahmanrashed (approved):

vava-odoo:
@robodoo r+

PR #2204, opened 2026-05-21
[FIX] 3pl_logistic_company: fix ownerless products on all users pages

frva-odoo:
This PR fixes the issue of products that have all requirements to be displayed on the website products page of the user but no assigned owners are displayed on the page of some owners.

Issue was coming from the value of parent_id being false for some partners, explaining why a Falsy owner was still displayed.

This also changes the use of x_owner_id from the product to using the owner_id
of the stock quant, so that the user view the quants he has no matter the initial
x_owner_id on the product.

TASK-6234967

### DIFF
--- a/3pl_logistic_company/data/portal_templates.xml
+++ b/3pl_logistic_company/data/portal_templates.xml
@@ -15,7 +15,7 @@
     <template id="portal_products" name="Your Products">
         <t t-call="portal.portal_layout">
             <t t-set="logged_user" t-value="request.env.user"/>
-            <t t-set="user_products" t-value="request.env['stock.quant'].sudo().search([('product_id.x_owner_id', 'in', [logged_user.partner_id.id, logged_user.partner_id.parent_id.id]), ('location_id.usage', '=', 'internal'), ('quantity', '!=', '0')])"/>
+            <t t-set="user_products" t-value="request.env['stock.quant'].sudo().search([('owner_id', '!=', False), ('owner_id', 'parent_of', logged_user.partner_id), ('location_id.usage', '=', 'internal'), ('quantity', '!=', '0')])"/>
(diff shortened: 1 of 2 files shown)


3pl_logistic_company/data/portal_templates.xml
```diff
@@ -15,7 +15,7 @@
     <template id="portal_products" name="Your Products">
         <t t-call="portal.portal_layout">
             <t t-set="logged_user" t-value="request.env.user"/>
-            <t t-set="user_products" t-value="request.env['stock.quant'].sudo().search([('product_id.x_owner_id', 'in', [logged_user.partner_id.id, logged_user.partner_id.parent_id.id]), ('location_id.usage', '=', 'internal'), ('quantity', '!=', '0')])"/>
+            <t t-set="user_products" t-value="request.env['stock.quant'].sudo().search([('owner_id', '!=', False), ('owner_id.id', 'in', [logged_user.partner_id.id, logged_user.partner_id.parent_id.id]), ('location_id.usage', '=', 'internal'), ('quantity', '!=', '0')])"/>
```
vava-odoo: couldn't we use parent_of here? Because the parent tree can be longer
Also, there seems to be quite a custom logic on product_id.x_owner_id, while there is none on the quant owner_id. You're sure about this change ? The commit message is only about falsy values, not this.

frva-odoo: I discussed with LPIE, the x_owner_id on the product is used to define a default owner_id on a stock_quant but that can be changed and should be viewed on the portal by the owner on the quant, not the owner on product. This was an edge case we overviewed until now, I will add it in the commit message.

I forgot we could have multiple parents, I will adapt it to take that into account

vava-odoo (approved):
@robodoo r+

PR #2208, opened 2026-05-22
[FIX] hotel: fix product reference

dhrs-odoo:
Before this PR in product_id we were referencing product template record id due to which it raised error
In this commit we are correctly referencing the product id reference


vava-odoo (approved):
@robodoo up to saas-19.2
@robodoo r+

vava-odoo:
@robodoo override=ci/industry

PR #2220, opened 2026-05-22
[FIX] test_booking_engine: assign calendar to overbooking test resource

chga-odoo:
Explicitly assign calendar_id as False to the test resource. Since this calendar keeps the resource available on all days of the week, the test no longer depends on the weekday it runs on.


vava-odoo:
@robodoo r+ fw=no
(fixed in 19.1)

PR #2222, opened 2026-05-22
[FIX] *: Simplify code by removing useless browse

frva-odoo:
This PR removes the useless wording of env.browse(ref().id) to use the reference directly instead.


frva-odoo:
Hello @vava-odoo, we talked about this a while ago but this wording is equivalent to the old one, using ref directly instead of browsing id of the ref. Small PR to remove the old one from industry.

vava-odoo (approved):

vava-odoo:
@robodoo r+

PR #2229, opened 2026-05-25
[IMP] real_estate: improve property appointment creation flow

dhgoy-odoo:
Appointment workflow:
- Add an "Appointments" smart button on the property form view.
- Display the total number of appointments linked to the property.
- Open the property's related appointments when clicking on the smart button.
- Redirect users to the created appointment records after clicking on the
"Create Visit Link" button.

Demo data and test update:
- Since the Agent field is mandatory for appointment creation, add x_agent_id in
the demo property records.
- Add x_agent_id in the property test data since the agent is required for visit
link creation.

Bug fix:

Issue:
- The "Create Appointment & Link" action was creating duplicate appointments
and links even for properties whose link was already generated.

Steps to reproduce:
- Open the property list view.
- Select properties whose appointments are already created, and a link is
generated.
- Click on Actions -> "Create Appointment & Link".

Solution:
- Skip properties whose link has already been generated when creating
appointments/link from the server action menu.
- Generate visit links and appointments only for selected properties whose links
are not already generated, and show a validation message when all selected
Properties already have existing links.

Task ID-6217777

### DIFF
--- a/real_estate/__manifest__.py
+++ b/real_estate/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Real Estate Agency',
-    'version': '1.9',
+    'version': '1.10',
     'category': 'Services',
     'author': 'Odoo S.A.',
     'depends': [
--- a/real_estate/data/ir_actions_server.xml
+++ b/real_estate/data/ir_actions_server.xml
@@ -106,15 +106,35 @@ for record in records: record['x_technical_partner_id'] = env['res.partner'].cre
           ]]></field>
     </record>
 
+    <record id="server_action_property_appointments" model="ir.actions.server">
+        <field name="name">Open Appointments</field>
+        <field name="model_id" ref="product.model_product_template"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+appointments = records.x_appointment_ids
+action = {
+    'type': 'ir.actions.act_window',
+    'res_model': 'appointment.type',
+}
+action.update({'view_mode': 'form', 'res_id': appointments.id}) if len(appointments) == 1 else action.update({'name': 'Appointments', 'view_mode': 'list,form', 'domain': [('id', 'in', appointments.ids)]})
+        ]]></field>
+    </record>
+
     <record id="create_appointment_link_server_action" model="ir.actions.server">
         <field name="name">Create Appointment &amp; Link</field>
         <field name="model_id" ref="product.model_product_template"/>
         <field name="binding_model_id" ref="product.model_product_template"/>
         <field name="state">code</field>
         <field name="code"><![CDATA[
+appointment_tz = env.ref('base.user_admin').tz or 'Europe/Brussels'
 for record in records:
+    if record.x_booking_link: continue
+    if not record.x_agent_id: raise UserError(f"Please assign an agent first.")
     link = env['appointment.invite'].create({'short_code': datetime.datetime.now().strftime("%M%m%S%d%Y%H") + str(record.id), 'display_name': record.name})
-    env['appointment.type'].create({'name': record.name, 'display_name': record.name, 'x_property_id': record.id, 'appointment_invite_ids': [Command.link(link.id)], 'location_id': record.x_technical_partner_id.id, 'image_1920': record.image_1920, 'appointment_tz': env.ref('base.user_admin').tz or 'Europe/Brussels', 'staff_user_ids': [Command.link(record.x_agent_id.id)] if record.x_agent_id else False, 'category': 'custom', 'event_videocall_source': False})
+    appointment = env['appointment.type'].create({'name': record.name, 'display_name': record.name, 'x_property_id': record.id, 'appointment_invite_ids': [Command.link(link.id)], 'location_id': record.x_technical_partner_id.id, 'image_1920': record.image_1920, 'appointment_tz': appointment_tz, 'staff_user_ids': [Command.link(record.x_agent_id.id)], 'category': 'custom', 'event_videocall_source': False})
+
+if env.context.get('from_form_view') and len(records) == 1:
+    action = {'type': 'ir.actions.act_window', 'res_model': 'appointment.type', 'view_mode': 'form', 'res_id': appointment.id}
         ]]></field>
     </record>
 
--- a/real_estate/data/ir_model_fields.xml
+++ b/real_estate/data/ir_model_fields.xml
@@ -322,6 +322,19 @@ for property in self: property['x_x_property_matches_contacts_res_partner_count'
         <field name="selectable" eval="False"/>
         <field name="store" eval="False"/>
     </record>
+    <record id="field_product_template_appointment_count" model="ir.model.fields">
+        <field name="compute"><![CDATA[
+for record in self:
+    record['x_appointment_count'] = len(record.x_appointment_ids)
+        ]]></field>
+        <field name="ttype">integer</field>
+        <field name="field_description">Appointment Count</field>
+        <field name="model_id" ref="product.model_product_template"/>
+        <field name="name">x_appointment_count</field>
+        <field name="readonly" eval="True"/>
+        <field name="selectable" eval="False"/>
+        <field name="store" eval="False"/>
+    </record>
     <record id="field_appointment_type_property_id" model="ir.model.fields">
         <field name="ttype">many2one</field>
         <field name="copied" eval="True"/>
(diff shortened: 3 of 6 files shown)


real_estate/data/ir_actions_server.xml
```diff
@@ -97,15 +97,41 @@ for record in records: record['x_technical_partner_id'] = env['res.partner'].cre
           ]]></field>
     </record>
 
+    <record id="action_open_appointments_server_action" model="ir.actions.server">
```
pgu-odoo: Improve the Id

real_estate/data/ir_actions_server.xml
```diff
@@ -97,15 +97,41 @@ for record in records: record['x_technical_partner_id'] = env['res.partner'].cre
...
+        else: raise UserError(f"Please assign an agent first.")
     link = env['appointment.invite'].create({'short_code': datetime.datetime.now().strftime("%M%m%S%d%Y%H") + str(record.id), 'display_name': record.name})
-    env['appointment.type'].create({'name': record.name, 'display_name': record.name, 'x_property_id': record.id, 'appointment_invite_ids': [Command.link(link.id)], 'location_id': record.x_technical_partner_id.id, 'image_1920': record.image_1920, 'appointment_tz': env.ref('base.user_admin').tz or 'Europe/Brussels', 'staff_user_ids': [Command.link(record.x_agent_id.id)] if record.x_agent_id else False, 'category': 'custom', 'event_videocall_source': False})
+    appointment = env['appointment.type'].create({'name': record.name, 'display_name': record.name, 'x_property_id': record.id, 'appointment_invite_ids': [Command.link(link.id)], 'location_id': record.x_technical_partner_id.id, 'image_1920': record.image_1920, 'appointment_tz': env.ref('base.user_admin').tz or 'Europe/Brussels', 'staff_user_ids': [Command.link(record.x_agent_id.id)], 'category': 'custom', 'event_videocall_source': False})
```
pgu-odoo: You can take these to out side of loop.

```
admin = env.ref('base.user_admin')
appointment_tz = admin.tz or 'Europe/Brussels'
```

pgu-odoo (commented):
@vava-odoo PR is ready for your review.
@dhgoy-odoo Good work. Just a few remarks.

real_estate/data/ir_actions_server.xml
```diff
@@ -97,15 +97,43 @@ for record in records: record['x_technical_partner_id'] = env['res.partner'].cre
...
+
     <record id="create_appointment_link_server_action" model="ir.actions.server">
         <field name="name">Create Appointment &amp; Link</field>
         <field name="model_id" ref="product.model_product_template"/>
         <field name="binding_model_id" ref="product.model_product_template"/>
         <field name="state">code</field>
         <field name="code"><![CDATA[
+appointment_ids = []
+admin = env.ref('base.user_admin')
+appointment_tz = admin.tz or 'Europe/Brussels'
```
vava-odoo: No need to add 2 lines for this, keep it as it was

real_estate/data/ir_actions_server.xml
```diff
@@ -97,15 +97,43 @@ for record in records: record['x_technical_partner_id'] = env['res.partner'].cre
...
 for record in records:
+    if record.x_booking_link: continue
+    if not record.x_agent_id:
+        if len(records) > 1: raise UserError(f"Please assign an agent in {record.name} property.")
+        else: raise UserError(f"Please assign an agent first.")
```
vava-odoo: Do we need to make a distinction here? We are on the form view, no need for the record name
```suggestion
        raise UserError("Please assign an agent first.")
```

real_estate/data/ir_actions_server.xml
```diff
@@ -97,15 +97,43 @@ for record in records: record['x_technical_partner_id'] = env['res.partner'].cre
...
     link = env['appointment.invite'].create({'short_code': datetime.datetime.now().strftime("%M%m%S%d%Y%H") + str(record.id), 'display_name': record.name})
-    env['appointment.type'].create({'name': record.name, 'display_name': record.name, 'x_property_id': record.id, 'appointment_invite_ids': [Command.link(link.id)], 'location_id': record.x_technical_partner_id.id, 'image_1920': record.image_1920, 'appointment_tz': env.ref('base.user_admin').tz or 'Europe/Brussels', 'staff_user_ids': [Command.link(record.x_agent_id.id)] if record.x_agent_id else False, 'category': 'custom', 'event_videocall_source': False})
+    appointment = env['appointment.type'].create({'name': record.name, 'display_name': record.name, 'x_property_id': record.id, 'appointment_invite_ids': [Command.link(link.id)], 'location_id': record.x_technical_partner_id.id, 'image_1920': record.image_1920, 'appointment_tz': appointment_tz, 'staff_user_ids': [Command.link(record.x_agent_id.id)], 'category': 'custom', 'event_videocall_source': False})
+    appointment_ids.append(appointment.id)
+if not appointment_ids: raise UserError("All selected properties already have appointments.")
```
vava-odoo: why did you add this part? the button is only on the form, so there should only be one record.
Maybe you need to update create_link.xml instead, to only do it on a single record at a time

real_estate/data/ir_actions_server.xml
dhgoy-odoo: I added this distinction because the "Create Appointment & Link" server action is also available from the list view Actions menu, where users can select multiple properties and execute it in batch.

In that case, including the property name in the error message helps identify which property is missing an agent. That's why I added this logic.

real_estate/data/ir_actions_server.xml
dhgoy-odoo: As mentioned above, this action can also be executed from the list view on multiple selected properties.
In that case, if all selected properties already have an appointment linked, this message provides useful feedback to the user.

real_estate/data/ir_actions_server.xml
```diff
@@ -97,15 +97,42 @@ for record in records: record['x_technical_partner_id'] = env['res.partner'].cre
...
+    </record>
+
     <record id="create_appointment_link_server_action" model="ir.actions.server">
         <field name="name">Create Appointment &amp; Link</field>
         <field name="model_id" ref="product.model_product_template"/>
         <field name="binding_model_id" ref="product.model_product_template"/>
         <field name="state">code</field>
         <field name="code"><![CDATA[
+appointment_ids = []
+appointment_tz = env.ref('base.user_admin').tz or 'Europe/Brussels'
```
vava-odoo, line 129: I still don't get why we need a new variable for this.

real_estate/data/ir_actions_server.xml
vava-odoo: Ok I forgot about the list view. But either you display all names at once, either you give the generic message without names

real_estate/data/ir_actions_server.xml
vava-odoo: Hmm, maybe be best to silently move on, no? No need for a usererror, it will just do nothing, but there is no error neither.
Did Loic ask to open only new appointments? Or all appointments linked to selected records? I'd go for second option, but maybe best to iterate with him

real_estate/__manifest__.py
```diff
@@ -1,6 +1,6 @@
 {
     'name': 'Real Estate Agency',
-    'version': '1.7',
+    'version': '1.8',
```
vava-odoo: this is obsolete, please update

real_estate/data/ir_actions_server.xml
dhgoy-odoo: Before, this expression was inside the loop, so `env.ref('base.user_admin').tz` was evaluated on every iteration. Since the value is the same for all records, I moved it outside the loop to avoid repeating the same lookup multiple times. That's why I introduced a new variable.

real_estate/data/ir_actions_server.xml
dhgoy-odoo: Ok, I'll use a generic message without the property name.

real_estate/data/ir_actions_server.xml
dhgoy-odoo: I discussed it with Loïc, and he suggested not opening any appointment when the action is executed from the list view. The appointment form is only opened when the action is triggered from the form view.

vava-odoo (approved):

vava-odoo:
@robodoo override=ci/industry
@robodoo r+

PR #2233, opened 2026-05-26
[FIX] test_booking_engine: set pickup/return times

abdrahmanrashed:
Before this commit there were not pickup/return times added for product in overbooking ecommerce tests, which caused the start/end times of the rental order to be different from the intended start/end times. After this commit, the pickup/return times have been set and the tests now check that the sale order time is exactly the same as the pickup/return times, instead of withing a relative possible timeframe as how it was.

### DIFF
--- a/tests/test_booking_engine/tests/test_overbooking.py
+++ b/tests/test_booking_engine/tests/test_overbooking.py
@@ -1,6 +1,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from datetime import datetime, timedelta
+from zoneinfo import ZoneInfo
 
 from odoo.addons.payment.tests.http_common import PaymentHttpCommon
 from odoo import Command
@@ -32,6 +33,8 @@ class TestOverbooking(PaymentHttpCommon):
             'sale_ok': True,
             'list_price': 10,
             'rent_periodicity': 'days',
+            'pickup_time': 18,
+            'return_time': 9,
             'is_published': True,
         })
         cls.express_checkout_billing_values = {
@@ -122,7 +125,7 @@ class TestOverbooking(PaymentHttpCommon):
         with self.assertRaises(ValidationError, msg="This Sales Order can't be confirmed. No resources are available for the shifts in: %s." % self.product.name):
             sale_order.action_confirm()
 
-    def _test_ecommerce_book_stay_offer_available(self):
+    def test_ecommerce_book_stay_offer_available(self):
         """
         Scenario: Customer from ecommerce can book a stay offer when resources are available
         """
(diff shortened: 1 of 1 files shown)


tests/test_booking_engine/tests/test_overbooking.py
```diff
@@ -136,6 +140,11 @@ def _test_ecommerce_book_stay_offer_available(self):
         })
         self.assertEqual(response['tracking_info'][0]['item_id'], self.product.id, "No item found in shopping cart.")
         sale_order = self.env['sale.order'].search([], limit=1)
+        website_tz = ZoneInfo(self.env.company.website_id.tz)
+        picking_time = datetime.combine(self.start_date, float_to_time(self.product.pickup_time)) - timedelta(seconds=website_tz.utcoffset(self.start_date).total_seconds())
```
frva-odoo: Seems like utcoffset() can return a day difference, which I'm pretty sure is not wanted here in the time argument of datetime.combine() (You could go over the limit and so be off by a day in the result)

Image

frva-odoo: Overall I find it pretty weird to use combine for a day and a time, and calculate the difference in timezone in seconds.
Wouldn't it be better if we changed the timezone, applied changes then come back ? That way we are sure to have the right time set in our db :

```suggestion
        picking_time = self.start_date.astimezone(website_tz).replace(hour=self.product.pickup_time).astimezone(UTC).replace(tzinfo=None)
```
What do you think about this approach, is this correctly what we want ?

frva-odoo (commented):
Hello 👋 Vava asked me to have a look as I'm working with timezones too lately, I have two comments about this. Not sure what the best way to do this but we could have a talk about it if needed.

tests/test_booking_engine/tests/test_overbooking.py
abdrahmanrashed: That doesn't matter because the pickup and return times are fixed in the test case

abdrahmanrashed: I did it like that to be coherent with the respective code in `_cart_add` that applied this logic.
Your suggestion is shorter, but honestly I am not sure if there is any functional difference between the 2 implementations in this case. I guess I will keep it as is unless there is a reason to change.

vava-odoo: I like the suggestion

vava-odoo (approved):
@delegate=frva-odoo
Thanks both

vava-odoo:
@robodoo delegate=frva-odoo
(sorry @DeleGate :see_no_evil: I need a coffee)

frva-odoo:
@robodoo r+

PR #2240, opened 2026-05-28
[IMP] custom_industrial_equipment: remove 'none' from `tracking` field

lugr-odoo:
The 'none' selection is removed from the `tracking` field. Thus,
`tracking = 'none'` is now represented as `tracking = False` and
`is_storable = True` at the same time.

Task ID: [6106458](https://www.odoo.com/odoo/my-tasks/6106458)


vava-odoo:
@robodoo override=ci/industry
@robodoo r+

PR #2241, opened 2026-05-28
[REV] booking_engine: slot kanban view extension

vava-odoo:
In a recent commit, a planning view (extension) was replaced by another one (primary). This was meant to not pollute the standard planning kanban view with a default_group_by. The problem is that other xpath should remain in all views, i.e. in extension mode.

This commit brings back the previous view with the adaptation done in here-under mentioned commit, and only leaves the default_group_by attribute in the new primary view. It also fixes the inherit_id of the new view, which relies on sale_planning extension.

https://github.com/odoo/industry/commit/5e0786cc729de81779d68e59108cfbcfbed3d913

### DIFF
--- a/booking_engine/__manifest__.py
+++ b/booking_engine/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Booking Engine',
-    'version': '1.14',
+    'version': '1.15',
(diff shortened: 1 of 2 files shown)


vava-odoo:
@robodoo r+

PR #2242, opened 2026-05-28
[IMP] excise_management: improve excise report

abdrahmanrashed:
- Added domain to Excise Reports to filter out excises without license;
- Ordered the stockpickings by Creation Dates;
- Added new excise types for scraping and mrp production;
- Changed base excise report line model from `stock.move` to
`stock.move.line` to include missing lines that were not
displayed before.

task-6070478


excise_management/demo/excise_report.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo noupdate="1">
+    <record id="excise_report_1" model="x_excise_report">
```
vava-odoo, line 3: why in data/ ? It was in demo, and is only used in demo

excise_management/data/ir_actions_server.xml
```diff
@@ -59,7 +59,7 @@ for idx, tax in enumerate(all_taxes, start=1):
         <field name="active" eval="True"/>
         <field name="nextcall" eval="(datetime.now() + timedelta(days=(0 - datetime.now().weekday()) % 7)).strftime('%Y-%m-%d %H:%M:%S')" />
         <field name="code"><![CDATA[
-excise_licenses = env['x_excise_license'].search([]).ids or [False]
+excise_licenses = env['x_excise_license'].search([]).ids + [False]
```
vava-odoo, line 62: Is it because x_excise_license_id was not set in the demo?

excise_management/data/ir_model_fields.xml
```diff
@@ -171,73 +171,72 @@ for template in self:
     <record id="stock_move_currency" model="ir.model.fields">
         <field name="ttype">many2one</field>
         <field name="field_description">Currency</field>
-        <field name="model_id" ref="stock.model_stock_move"/>
+        <field name="model_id" ref="stock.model_stock_move_line"/>
```
vava-odoo, line 174: better change the id as well

excise_management/data/ir_model_fields.xml
```diff
@@ -619,12 +628,12 @@ for record in self:
         <field name="name">x_color</field>
         <field name="ttype">integer</field>
         <field name="field_description">Color</field>
-        <field name="model_id" ref="stock.model_stock_move"/>
+        <field name="model_id" ref="stock.model_stock_move_line"/>
```
vava-odoo, line 631: same

excise_management/data/ir_model_fields.xml
```diff
@@ -475,7 +474,7 @@ for record in self:
         <field name="name">x_currency_id</field>
         <field name="readonly" eval="True"/>
         <field name="relation">res.currency</field>
-        <field name="related">x_move_ids.product_id.currency_id</field>
+        <field name="related">x_line_ids.move_id.product_id.currency_id</field>
```
vava-odoo, line 477: I think it would be better to change the ID of this one as well

excise_management/data/ir_ui_view.xml
```diff
@@ -90,21 +89,20 @@
...
-        <field name="model">stock.move</field>
+    <record id="stock_line_view_list" model="ir.ui.view">
+        <field name="name">stock.move.line.view.list.excise</field>
+        <field name="model">stock.move.line</field>
         <field name="priority">99</field>
         <field name="active" eval="True"/>
         <field name="type">list</field>
         <field name="arch" type="xml">
-            <list default_order="date, picking_id, sequence" string="Moves" create="0">
+            <list default_order="date, picking_id" string="Moves" create="0">
```
vava-odoo, line 99: Moves?

excise_management/data/ir_ui_view.xml
```diff
@@ -314,4 +314,17 @@
             </list>
         </field>
     </record>
+    <record id="stock_vpicktree_inherit" model="ir.ui.view">
+        <field name="inherit_id" ref="stock.vpicktree"/>
+        <field name="mode">extension</field>
+        <field name="model">stock.picking</field>
+        <field name="name">stock.picking.list.inherit.vineyard</field>
```
vava-odoo: rename

excise_management/__manifest__.py
```diff
@@ -13,6 +13,7 @@
     'data': [
         'data/ir_model.xml',
         'data/ir_model_fields.xml',
+        'data/excise_report.xml',
```
vava-odoo: version bump required

beverage_distributor/demo/stock_move.xml
vava-odoo, line 1: why do you need to create the stock.move(.line) records? Can't you update (write) the necessary information based on the linked purchase orders? Same for pickings

micro_brewery/demo/stock_warehouse.xml
```diff
@@ -1,10 +1,15 @@
...
-    </record>
     <record id="warehouse_1" model="stock.warehouse">
         <field name="name">Bar</field>
         <field name="code">Bar</field>
     </record>
+    <record id="stock.warehouse0" model="stock.warehouse" forcecreate="0">
+        <field name="sequence">1</field>
+        <field name="x_is_fiscal_deposit" eval="True"/>
+    </record>
+    <record id="excise_management.stock_warehouse_fiscal_deposit" model="stock.warehouse" forcecreate="0">
```
vava-odoo, line 11: you don't change anything here from what is defined in excise, do you?

vineyard/demo/sale_order.xml
vava-odoo, line 1: I don't get why you do this here

excise_management/data/ir_actions_server.xml
abdrahmanrashed: we can have reports without ids, without this change we would miss them if we have licenses

excise_management/data/ir_ui_view.xml
abdrahmanrashed: it is called like that in standard:

Image

micro_brewery/demo/stock_warehouse.xml
abdrahmanrashed: the sequence gets overriden so we have to redefine it here
(will remove the other field)

vineyard/demo/sale_order.xml
abdrahmanrashed: request from @Pierre-savp to have the dates be from this year instead of last year

vava-odoo: I meant the unlink/cancel part

excise_management/data/ir_actions_server.xml
vava-odoo: I don't get that. Shouldn't we enforce licence?

abdrahmanrashed: if the business has only 1 license they would need to write it everywhere hence the option of it not being enforced.

so from a functional perspective if they dont have a license on a report all reports wont have a license (hence the original implementation as well), however a business started having multiple licenses then there would be a point when some reports would have a licenses written and others not (and the ones without wouldnt be displayed before the change).

the concern here is to minimize issues with the functionality, perhaps there could be better ways to assign licenses but for now this is left for businesses to manage themselves.

vineyard/demo/sale_order.xml
abdrahmanrashed: there are sale orders automatically created idk how and this is the only way i found to remove them, you cant unlink them without cancelling them and the analytics depend on them so they prevent you to unlink as well

vava-odoo: better investigate where they come from first, as there is supposedly no other demo data than industry ones

beverage_distributor/demo/stock_move.xml
abdrahmanrashed: turns out they were duplicated and we didnt even need them (should have checked my bad)

vineyard/demo/sale_order.xml
abdrahmanrashed: turns out they are related to other stuff which would require a moderate change to the module unrelated to the scope of the current task. a new task will be created for this

vava-odoo (approved):
@robodoo r+ squash

PR #2251, opened 2026-06-01
[IMP] *: Add name and logo for main company, reword mails

frva-odoo:
This commit touches almost all industries, it first add the demo/res_company.xml file to industries missing it and use it then to give the name "My[industry_name]" to the main_company in all industries. It then gives the logo of the industry to the company. At last, it renames all emails found at the format "...@[company_name].example.com" through all files in all industries.

This was made by using bash scripts, these can be found on the task.

Task-6180590


tests/test_generic/tests/test_xml.py
```diff
@@ -108,6 +108,17 @@
     "filter_pre_domain",
 }
 
+TECHNICAL_MODULES = [
```
vava-odoo: why not use the difference between installed_modules and installed_industries?

tests/test_generic/tests/test_xml.py
```diff
@@ -785,3 +799,28 @@ def _check_base_records_update(self, root, file_name, module):
                             file_name,
                             field_name,
                         )
+
+    def _check_main_company_inherit_is_present(self, root, files, module):
+        if module in TECHNICAL_MODULES:
+            return
+        for file in files:
```
vava-odoo, line 794: the main method already loops over files, why do you need to it again here?

textile_manufacturing/demo/res_company.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo noupdate="1">
+    <record id="base.main_company" model="res.company" forcecreate="1">
```
vava-odoo: just wondering: if you create a second company, then removes the first one. Is this xmlid linked to the company left?
What I had in mind is that maybe forcecreate could be False here

3pl_logistic_company/demo/res_partner.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo noupdate="1">
     <record id="base_industry_data.res_partner_32" model="res.partner" forcecreate="0">
-        <field name="email">second_company@yourcompany.example.com</field>
+        <field name="email">second_company@MyLogisticCompany.example.com</field>
```
vava-odoo, line 4: I remember we discussed about portal ones. We don't apply to them?

vava-odoo (approved):
Just a few comments

tests/test_generic/tests/test_xml.py
frva-odoo: Here I want to check if any file in the demo folder contains base.main_company, and I have to throw a warning if none contains it. If I wanted to put this in the file loops in _check_files_in_path, I would have to put the logic of "if not found raise warning" after the loop and I think it is messy and unclear so I put it separately in this method executed after instead.

3pl_logistic_company/demo/res_partner.xml
frva-odoo: From SAVP: No need

vava-odoo:
@robodoo r+

PR #2253, opened 2026-06-01
[FIX] condominium, property_assets_distribution: fix error on property deletion

assh-odoo:
Currently, an error occurs when a user deletes a property record.

**Steps to Reproduce:**
 - Install `industry_real_estate` with demo data.
 - Go to `Leasing` > `Properties` > `Properties`.
 - Delete a `property` record.

`ValueError: AttributeError("'x_property' object has no attribute 'x_owner_ids'") while evaluating
"record.x_owner_ids.x_account.write({'active': False})"`

After a [recent commit], the x_owner_ids field was defined on x_property in the condominium 
module [1]. The error occurs when a property record is deleted and a server action [2] is triggered 
through a base automation rule [3] to archive the owners' account records. Since x_owner_ids is 
defined in the condominium module, accessing it raises an error when the condominium module 
is not installed.

This commit moves the server action and the corresponding base automation rule to the 
condominium module, where the x_owner_ids field is defined.

[recent commit]: https://github.com/odoo/industry/commit/8e166b3bb8bb66e53da5c2f9e6e15a9751f7aa38#diff-70ae8e6ca07f28b93f3e4d02c124d0217e4e951b2b5f77a92cc9651142c22c0d

[1]: https://github.com/odoo/industry/blob/d0b2141b47115dc8dd7a5e5fca4131a227125435/condominium/data/ir_model_fields.xml#L200-L208

[2]- https://github.com/odoo/industry/blob/d0b2141b47115dc8dd7a5e5fca4131a227125435/property_assets_distribution/data/ir_actions_server.xml#L47-L55

[3]: https://github.com/odoo/industry/blob/d0b2141b47115dc8dd7a5e5fca4131a227125435/property_assets_distribution/data/base_automation.xml#L10-L15

sentry-7511045457

### DIFF
--- a/condominium/__manifest__.py
+++ b/condominium/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Property Owner Association',
-    'version': '3.1',
+    'version': '3.2',
     'category': 'Services',
     'depends': [
         'account_check_printing',
--- a/condominium/data/base_automation.xml
+++ b/condominium/data/base_automation.xml
@@ -37,4 +37,10 @@
         <field name="action_server_ids" eval="[(6, 0, [ref('action_update_voter_line_for_motion')])]"/>
         <field name="trigger_field_ids" eval="[(6, 0, [ref('field_calendar_attendee_is_attending'), ref('field_calendar_attendee_is_delegating')])]"/>
     </record>
+    <record id="automation_archive_related_accounts" model="base.automation">
+        <field name="name">On property deletion</field>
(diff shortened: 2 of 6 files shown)


maan-odoo (approved):

abdrahmanrashed (approved):

vava-odoo (approved):
Thanks for the fix!
Could you just rebase (and adapt version of condominium)?

vava-odoo:
@robodoo r+

PR #2257, opened 2026-06-02
[IMP] textile_manufacturing: make knowledge demo data portable

aycha-odoo:
- The article bodies had hardcoded database IDs in their embedded view props, which would break on any fresh install since those IDs are auto-assigned by Postgres and differ every time.

- Switch body fields to type=html and use %(module.xml_id)d so Odoo resolves the correct IDs at install time.

- Reorder records so children exist before parent bodies reference them.

- Also add the two design images as proper module attachments instead of URLs.

Task-6263926


vava-odoo:
Already done with https://github.com/odoo/industry/commit/d3bf0c1b8f1c15356b67e175af19400caf75d24e

PR #2262, opened 2026-06-02
[IMP] textile_manufacturing: make knowledge demo data portable

aycha-odoo:
- The article bodies had hardcoded database IDs in their embedded view props, which would break on any fresh install since those IDs are auto-assigned by Postgres and differ every time.

- Switch body fields to type=html and use %(module.xml_id)d so Odoo resolves the correct IDs at install time.

- Reorder records so children exist before parent bodies reference them.

- Also add the two design images as proper module attachments instead of URLs.

task-6263926


tests/test_generic/tests/test_xml.py
```diff
@@ -389,12 +389,18 @@ def _check_knowledge_article_is_published(self, root, file_name):
...
+            if record_id := record.get("id"):
+                all_ids.add(record_id)
+                if record.xpath(".//field[@name='is_locked']/@eval") == ["True"]:
+                    locked_ids.add(record_id)
+        for record_id in all_ids - locked_ids:
+            _logger.warning(
+                "Knowledge article %s in %s should have 'is_locked' set to True.",
+                record_id,
+                file_name,
+            )
```
pgu-odoo, line 403: why sets are needed?

will this only not work?

```
for record in records:
    if record_id := record.get("id"):
        if record.xpath(".//field[@name='is_locked']/@eval") != ["True"]:
            _logger.warning(
                "Knowledge article %s in %s should have 'is_locked' set to True.",
                record_id,
                file_name,
            )
```

pgu-odoo (commented):
@aycha-odoo Good work! Just a small feedback.

tests/test_generic/tests/test_xml.py
aycha-odoo: That was actually my initial approach. While testing, I realized that the requirement is satisfied if at least one knowledge article record has is_locked="True".

For example, if a file contains two knowledge articles and the first one does not have is_locked="True" but the second one does, the test should pass. With the per-record validation, the first article would still be flagged even though the file already contains a valid locked article. A bit unfair to the second article, which was doing its job. 😄

That's why I changed the logic to check whether any knowledge article in the file has is_locked="True" rather than requiring every record to have it.

tests/test_generic/tests/test_xml.py
```diff
@@ -389,12 +389,18 @@ def _check_knowledge_article_is_published(self, root, file_name):
...
+        all_ids, locked_ids = set(), set()
         for record in root.xpath("//record[@model='knowledge.article']"):
-            is_locked_fields = record.xpath(".//field[@name='is_locked']/@eval")
-            if not is_locked_fields:
-                _logger.warning(
-                    f"Knowledge article in {file_name} should have 'is_locked' set to True."
-                )
+            if record_id := record.get("id"):
+                all_ids.add(record_id)
+                if record.xpath(".//field[@name='is_locked']/@eval") == ["True"]:
```
vava-odoo: Couldn't you be more permissive? All truthy values should be ok here, especially 1

vava-odoo (approved):

vava-odoo:
@robodoo r+ rebase-ff

PR #2265, opened 2026-06-02
[FIX] construction: remove color from resource

kthe-odoo:
In the PR https://github.com/odoo/odoo/pull/267666 and https://github.com/odoo/enterprise/pull/119080 , we have removed colors on resources and they will use colors from roles set on them.

task-4678837


vava-odoo (approved):
@robodoo override=ci/industry
@robodoo r+

PR #2270, opened 2026-06-02
[FIX] real_estate: create ribbons to avoid translations issues

vava-odoo:
[FIX] real_estate: create ribbons to avoid translations issues

Before this commit, the standard ecommerce ribbons were renamed in the
industry module. While this works fine in English, the translation
mechanism will only search for translated terms from the standard
module. In such a case, the industry term and the translated ones can
have very different meanings.

This commit creates new ribbons instead, and unlinks standard ones (only
if they are not linked to any product yet).

[FIX] wine_merchant: new attribute, no renaming of standard records

Before this commit, the standard product attributes were renamed in the
industry module. While this works fine in English, the translation
mechanism will only search for translated terms from the standard
module. In such a case, the industry term and the translated ones can
have very different meanings.

This commit creates new attributes instead.

task-6003598

### DIFF
--- a/real_estate/data/product_ribbon.xml
+++ b/real_estate/data/product_ribbon.xml
@@ -1,28 +1,39 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo noupdate="1">
-    <record id="website_sale.sale_ribbon" model="product.ribbon" forcecreate="1">
+    <record id="draft_ribbon" model="product.ribbon">
         <field name="name">Draft</field>
-        <field name="text_color">#ffffff</field>
         <field name="x_is_draft_ribbon" eval="True"/>
+        <field name="sequence">1</field>
     </record>
-    <record id="website_sale.sold_out_ribbon" model="product.ribbon" forcecreate="1">
+    <record id="pre_published_ribbon" model="product.ribbon">
         <field name="name">Pre-published</field>
         <field name="bg_color">#a3009e</field>
+        <field name="sequence">0</field>
     </record>
     <record id="website_sale.new_ribbon" model="product.ribbon" forcecreate="1">
-        <field name="name">New</field>
-        <field name="bg_color">#0275d8</field>
+        <field name="name">New!</field>
+        <field name="sequence">2</field>
     </record>
     <record id="product_ribbon_6" model="product.ribbon">
         <field name="name">Option</field>
         <field name="bg_color">#e0ce00</field>
+        <field name="sequence">3</field>
     </record>
     <record id="product_ribbon_7" model="product.ribbon">
         <field name="name">Sold</field>
         <field name="bg_color">#41a300</field>
         <field name="x_is_closed_ribbon" eval="True"/>
+        <field name="sequence">4</field>
     </record>
     <function model="product.ribbon" name="unlink">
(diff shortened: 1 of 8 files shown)


real_estate/data/product_ribbon.xml
```diff
@@ -1,28 +1,41 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo noupdate="1">
-    <record id="website_sale.sale_ribbon" model="product.ribbon" forcecreate="1">
+    <record id="draft_ribbon" model="product.ribbon">
         <field name="name">Draft</field>
         <field name="text_color">#ffffff</field>
```
trcazier: (Nitpicking) This is the default value of this field


```suggestion
```

real_estate/data/product_ribbon.xml
```diff
@@ -1,28 +1,41 @@
...
     </record>
-    <record id="website_sale.sold_out_ribbon" model="product.ribbon" forcecreate="1">
+    <record id="pre_published_ribbon" model="product.ribbon">
         <field name="name">Pre-published</field>
         <field name="bg_color">#a3009e</field>
+        <field name="sequence">0</field>
     </record>
     <record id="website_sale.new_ribbon" model="product.ribbon" forcecreate="1">
         <field name="name">New</field>
         <field name="bg_color">#0275d8</field>
```
trcazier: (Nitpicking) This is already in the original record


```suggestion
```

wine_merchant/data/product_attribute.xml
```diff
@@ -1,33 +1,42 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo noupdate="1">
-  <record id="website_sale.product_attribute_brand" model="product.attribute" forcecreate="1">
+  <record id="product_attribute_category" model="product.attribute">
     <field name="name">Product Category</field>
   </record>
   <record id="product_barcodelookup.product_attribute_lookup_1" model="product.attribute" forcecreate="1">
+    <field name="name">Color</field>
+  </record>
```
trcazier, line 8: I thought the whole point was to make new ones when we changed the name, unless there's something I'm missing?
Plus, this is already the name of the original record (minus the capital letter)
```suggestion
```

or

```suggestion
  <record id="product_barcodelookup.product_attribute_lookup_1" model="product.attribute" forcecreate="0">
    <field name="visibility" eval="False"/>
  </record>
  <record id="product_attribute_color" model="product.attribute">
    <field name="name">Color</field>
  </record>
```

trcazier (commented):
Overall LGTM except the Color one, I just have mixed feelings about entirely removing the `website_sale` ribbons which might break other flows but if BOJE said so... 😇

trcazier:
By the way, if you want to fix every instance of this 'bug', you will have some serious work to do

image

wine_merchant/data/product_attribute.xml
vava-odoo: > I thought the whole point was to make new ones when we changed the name

Right. And this one already exists, so we use it. If we create a new one, despite the visibility to False, it will create a duplicate in the backend.

trcazier (approved):
LGTM 🚀

vava-odoo:
@robodoo r+ rebase-ff

vava-odoo:
@robodoo override=ci/industry

PR #2276, opened 2026-06-05
[FIX] real_estate: fix automation to link mail reference to properties

vava-odoo:
Before this commit, this automation was not working for several reasons. The first and major one is that the domain on references was the inverse of the expected behaviour: it looks for the content or subject of the email to be contained in one of the reference. The second issue is that has_message was not False when a lead is created, and therefore this automation was never triggered.

This commit fixes both issues by editing the domain triggering the action and by looking at the db level for the reference that could be included in the email.

opw-6272644

### DIFF
--- a/real_estate/__manifest__.py
+++ b/real_estate/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Real Estate Agency',
-    'version': '1.4',
+    'version': '1.5',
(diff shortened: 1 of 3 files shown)


abdrahmanrashed (approved):

vava-odoo:
@robodoo r+

PR #2280, opened 2026-06-08
[FIX] excise_management: prevent reporter value from being reset to default

dhgoy-odoo:
Currently, the Reporter setting is reset to Administrator after saving the 
Inventory settings.

Step to reproduce:
* Go to Inventory → Configuration → Settings.
* Change the Reporter field to another user.
* Save the settings.

Cause:
The Reporter field relies on an `ir.default` record to provide its default value.
However, the corresponding `ir.default` entry was not associated with a specific
company. As a result, when the settings were saved, Odoo reapplied the global
default value stored in `ir.default`, overriding the user-selected Reporter and
resetting the field to Administrator.

Solution:
Add the `res_company` model's field to the corresponding `ir.default` record so that 
the default value is applied in the correct company context and does not overwrite
the Reporter value saved by the user.

Task ID:6283344

### DIFF
--- a/excise_management/__manifest__.py
+++ b/excise_management/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Excise Management',
-    'version': '2.7',
+    'version': '2.8',
     'category': 'Inventory/Inventory',
     'author': 'Odoo S.A.',
     'depends': [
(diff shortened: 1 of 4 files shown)


vava-odoo (approved):
@robodoo r+

PR #2301, opened 2026-06-11
[FIX] booking_engine: prevent Owl error in hotel planning kanban

dhvag-odoo:
**Steps to Reproduce:**
1. Go to Apps → Industry → Install Hotel (load demo data).
2. Open Hotel App.
3. Switch to the Kanban view.

**Result:**
image

**Cause of the Issue:**

- The sale_planning module adds the [`sale_line_id`](https://github.com/odoo/enterprise/blob/saas-19.2/sale_planning/views/planning_slot_views.xml#L143) field.

- Then, the booking_engine module removes the [`sale_line_id`](https://github.com/odoo/industry/blob/19.0/booking_engine/data/ir_ui_view.xml#L258) field from the kanban field declarations.

- However, the [`sale section`](https://github.com/odoo/enterprise/blob/saas-19.2/sale_planning/views/planning_slot_views.xml#L146) remains in the final template and still references: `record.sale_line_id.raw_value`

**As a result:**
Since `sale_line_id` is no longer loaded in the kanban record payload, the template attempts to access `record.sale_line_id.raw_value` on an undefined value, resulting in an OwlError.

**With this commit:**
Remove the corresponding sale section when removing `sale_line_id`. This keeps the kanban template consistent with the loaded fields and prevents the OwlError.

opw-6261815

### DIFF
--- a/booking_engine/__manifest__.py
+++ b/booking_engine/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Booking Engine',
-    'version': '1.18',
+    'version': '1.19',
     'category': 'Hidden/Tools',
     'author': 'Odoo S.A.',
     'depends': [
--- a/booking_engine/data/ir_ui_view.xml
+++ b/booking_engine/data/ir_ui_view.xml
@@ -250,6 +250,7 @@
     <field name="active" eval="True"/>
     <field name="arch" type="xml">
       <xpath expr="//field[@name='sale_line_id']" position="replace"/>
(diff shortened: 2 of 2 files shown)


dhrs-odoo (approved):
Hello @dhvag-odoo 
Thanks for the fix 🙌

vava-odoo (approved):
It's not the bug I wanted to solve, but this is quite annoying as well 😅 
Thanks for the fix. Could you just increase the minor version?

vava-odoo:
@robodoo r+

PR #2303, opened 2026-06-11
[FIX] base_industry_data: remove dependency to digest

vava-odoo:
In [this commit], the code to deactivate the digest cron in demo data of industry moved from a setting to archiving the cron. By doing so, the dependency to digest is not useful anymore, as it unarchives the cron only if it exists.

[this commit]: https://github.com/odoo/industry/commit/b7be437b9486030a526bfa05dbc5ab339446322a

### DIFF
--- a/base_industry_data/__manifest__.py
+++ b/base_industry_data/__manifest__.py
@@ -5,7 +5,6 @@
     'category': 'Hidden/Tools',
     'depends': [
         'base',
-        'digest',
         'knowledge',
     ],
     'data': [


amas-odoo:
@vava-odoo we can remove digest from the dependencies. Since digest is already included through the Knowledge module's dependencies, removing it will not have any functional impact and will only eliminate a redundant dependency.

ref: https://github.com/odoo/enterprise/blob/19.0/knowledge/__manifest__.py#L12

vava-odoo:
@amas-odoo I had hopes to reach a point where users could remove this module, but I guess we'll never get there. Well, at least, it's not our fault anymore :innocent: 
@robodoo r+

PR #2324, opened 2026-06-15
[FIX] headhunter: avoid duplicate document creation on demo data

amas-odoo:
The headhunter demo data was linking an attachment already associated with a document record. During applicant creation, Documents attempts to create a document for the attachment, resulting in:

    psycopg2.errors.UniqueViolation:
    duplicate key value violates unique constraint
    "documents_document_attachment_unique"

This behavior was introduced by:
https://github.com/odoo/enterprise/pull/112993

Fix by creating dedicated attachment records for the demo applicants instead of reusing existing attachments already linked to documents.


vava-odoo (approved):
@robodoo r+

PR #2333, opened 2026-06-16
[FIX] 3pl_logistic_company: fix portal products no access rights

frva-odoo:
This commit fixes a new "no access rights" error on the products portal page. This is caused by the recent change in standard from the commit odoo/odoo@846eb51 that changed how important was the model of a server action used by an url. Now we need the write accesses on the mode for triggering a server action, or as we do here have the access through the groups field of the server action.

Task-6306283


vava-odoo (approved):
@robodoo r+

PR #2335, opened 2026-06-16
[FIX] 3pl, headhunter: fix portal no access rights and add tests

frva-odoo:
This PR fixes a "no access rights" error on the portal page. This is caused by the recent change in standard from the commit odoo/odoo@846eb51 that changed how important was the model of a server action used by an url. Now we need the write accesses on the model for triggering a server action, or as we do here have the access through the groups field of the server action.

This also add tests so we can detect when it happens from now on.

Task-6306283


vava-odoo (approved):
@robodoo r+ rebase-ff

vava-odoo:
@robodoo override=ci/industry

PR #2354, opened 2026-06-18
[FIX] test_industry: fix the size of the registry

vava-odoo:
Since [this commit] it is not possible to define `registry_lru_size` from
the config. This commit fixes the issue by defining it in the os.environ
directly.

[this commit]: https://github.com/odoo/odoo/commit/7ec7048fef7f8cb6b0eb320903d6197c9841fec8


vava-odoo:
@robodoo r+ squash

PR #2357, opened 2026-06-18
[IMP] 3pl_logistic_company: add test for url and fix missing id

frva-odoo:
This commit adds a test for the portal page products, to be sure it is not accessible when not connected but accessible for any user with portal access.

This also fixes a missing id in a search.

Task-6306283


vava-odoo:
@robodoo up to saas-19.2

vava-odoo:
@frva-odoo runbot is red here

vava-odoo (approved):
@robodoo r+

PR #2378, opened 2026-06-22
[IMP] excise_management: add sale order excise tax scenarios

dhgoy-odoo:
These tests ensure that excise taxes are correctly applied or excluded based on the customer's fiscal position.

- Add a test for customers without a fiscal deposit:
  - Verify excise taxes are applied.
  - Verify the sale order line amount matches the unit price multiplied by the quantity.
  - Verify the tax-included amount equals the amount plus VAT taxes.

- Add a test for customers with a fiscal deposit:
  - Verify excise taxes are not applied.
  - Verify the sale order line amount is computed according to the fiscal position configuration.
  - Verify the tax-included amount equals the amount plus VAT taxes.

TaskId-6053673

### DIFF
--- a/tests/test_excise_management/tests/test_action_server.py
+++ b/tests/test_excise_management/tests/test_action_server.py
@@ -258,3 +258,104 @@ class ActionServerTestCase(TransactionCase):
         server_action.with_context(active_ids=[excise_report_2.id], active_model="x_excise_report").sudo().run()
         line_2 = self.env['x_excise_report_line'].search([('x_excise_report_id', '=', excise_report_2.id)])
         self.assertEqual(line_2.filtered(lambda x: diff_license_diff_warehouse_move in x.x_move_ids).x_excise_move_type, 'entry fd')
+
+    def test_sale_order_customer_without_fiscal_deposit(self):
+        customer = self.env['res.partner'].create({
+            'name': 'Customer Without Deposit',
+        })
+        template = self.env['product.template'].create({
+            'name': 'Product With Excise',
+            'list_price': 100,
+        })
+
+        with Form(template) as form:
+            form.x_excise_category = self.env.ref(
+                'excise_management.x_excise_category_S001'
+            )
+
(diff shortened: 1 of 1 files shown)


tests/test_excise_management/tests/test_action_server.py
```diff
@@ -258,3 +258,104 @@ def test_fiscal_deposit_move_computation(self):
...
+        with Form(order) as order_form:
+            with order_form.order_line.new() as line_form:
+                line_form.product_id = product
+                line_form.product_uom_qty = 1
+
+        order_form.save()
+        line = order.order_line
+
+        excise_taxes = line.tax_ids.filtered(
+            lambda t: t.tax_group_id.x_is_excise and t.name != 'No Excises'
```
vava-odoo: I don't think you need to filter the "No Excise". The goal is that it appears on the SO, but the amount of excise taxes should be 0.

vava-odoo (approved):
Thanks. Note that your commit should be an IMP (not an ADD).

tests/test_excise_management/tests/test_action_server.py
dhgoy-odoo: If I do this, then I have to check the excise amount instead:
```
self.assertEqual(excise_amount,  0,
    "Customer with fiscal deposit should not have excise taxes applied")
```
Or, if I should validate it using `excise_taxes`, then I would check:
```
self.assertTrue(not excise_taxes or excise_taxes.mapped('name') == ['No Excises'],
"Customer with fiscal deposit should not have excise taxes applied")
```
I think we should check the excise amount instead, since the excise_tax amount is always 0 in this situation. Is that OK?

vava-odoo: yeah, only the amount matters here

dhgoy-odoo: ok thanks

vava-odoo:
@robodoo r+

PR #2381, opened 2026-06-23
[REM] booking_engine: unenforce booking date hours

abdrahmanrashed:
This commit removes the server_action in `booking_engine` that used to force the dates for any stay offer booking to start at product `pickup_time` and end at product `return_time`.

task-6314477


vava-odoo:
@robodoo r+

PR #2383, opened 2026-06-23
[IMP] tests: allow other structure than data/demo

trcazier:
Before this commit, all manifest files had to fall either into the data or demo folder, or the tests would fail. This commit changes this by allowing any file listed in the manifest to have any path, allowing for a more flexible structure.

Note that this removes the ability to check for files that are detected but not listed in the manifest since they can be under either key ('data' or 'demo') and would be flagged as missing. Therefore, this was moved to a new check, _test_no_orphaned_files.

Part of task-6251595

### DIFF
--- a/tests/test_generic/tests/test_manifest.py
+++ b/tests/test_generic/tests/test_manifest.py
@@ -67,8 +67,6 @@ class ManifestTest(ManifestLinter, IndustryCase):
                 if key == 'license' and not any(mod in manifest_data.get('depends', []) for mod in ['base_industry_data', 'knowledge', 'web_studio']):
                     expected_value = 'LGPL-3'
                 self.assertEqual(value, expected_value, f"Wrong {key} '{value}' in manifest, it should be {expected_value}")
-            if key in ['data', 'demo']:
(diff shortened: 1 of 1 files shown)


tests/test_generic/tests/test_manifest.py
```diff
@@ -100,30 +99,37 @@ def _validate_assets(self, module, manifest_data):
...
+            _logger.warning("Duplicated in %s in the manifest: %s", manifest_key, ', '.join(duplicates))
+
+        # Check if the files listed in the manifest actually exist in the module
+        if (not_found := {file_path for file_path in manifest_files if not (self.module_path / file_path).is_file()}):
+            _logger.error("Files listed in manifest %s not found: %s", manifest_key, ', '.join(not_found))
+
+        # Flexible website.xml check (path-agnostic)
+        website_xml_files = [f for f in manifest_files_list if f.endswith('website.xml')]
+        if website_xml_files:
+            if not manifest_files_list[-1].endswith('website.xml'):
```
vava-odoo: this condition should probably be smoothed. It is only true when loading a theme. But with the new tasks coming from designers, they don't install any, so this constrain makes no sense

vava-odoo:
@robodoo r+ rebase-ff

PR #2403, opened 2026-06-24
[FIX] industry_real_estate: fix traceback when saving custom Contact us fields

dhrs-odoo:
Before this PR, adding a field to the Contact Us form through the website editor and saving the page would trigger a traceback.

The issue was caused by an incorrect input name in the form. The field was defined as `property` instead of the actual field name **`x_property_id`,** causing the whitelisting validation to fail with: Unable to whitelist field(s) ['Property']

This PR updates the input name to use the correct field, allowing the form to be saved successfully.

Task-6323839

### DIFF
--- a/industry_real_estate/__manifest__.py
+++ b/industry_real_estate/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Property Management',
-    'version': '1.3',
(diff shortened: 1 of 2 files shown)


vava-odoo:
@robodoo r+

PR #2407, opened 2026-06-24
[IMP] deposit_management: Full multi-company support

jowes-odoo:
Adds full multi-company support for the automation "Automated BoM." This includes creating a BoM and orderpoint(s) with the constraint of both products needing to be visible to all or visible to the same selected company.

Creating a BoM:
Only one BoM is ever created. The selected company will match the products. If the products are visible to all, so is the BoM. If the products have a company selected, so will the BoM.

Creating reordering rules (orderpoint):
Orderpoints are connected with a location, thus requiring one company to be selected. We create orderpoint records for each warehouse with a manufacturing route setup. This applies to all companies if the products are visible to all, or a single company if one is selected. The automation rule now dynamically determines the route based on the warehouse.

Changing the containing product:
If a user initially had Product A containing 10 Product Bs, but changes Product A to now contain 10 Product Cs, we need to create the BoM and orderpoints for Product C. However, we don't want to delete the records for Product B, as these may have already been used. If the user decides to go back to Product B, the BoM and orderpoints will already be made.

closes #2407
opw-6298628

### DIFF
--- a/deposit_management/data/ir_actions_server.xml
+++ b/deposit_management/data/ir_actions_server.xml
@@ -2,48 +2,56 @@
 <odoo>
     <record id="bom_server_action" model="ir.actions.server">
         <field name="code"><![CDATA[
-if (quant := record.x_quantity_by_deposit_product) == 1 and record.x_unit_sale_product.id:
+if (quant := record.x_quantity_by_deposit_product) == 1:
     raise UserError("Unit sale product should be undefined if this product can not be sold in a smaller unit.")
-if quant <= 0 and record.x_unit_sale_product: raise UserError("The auto-BOM should contain a positive amount of units.")
-if record.x_unit_sale_product:
-    existing_bom = env['mrp.bom'].search([('product_tmpl_id', '=',record.x_unit_sale_product.id)], limit=1)
-    reordering_rule = env['stock.warehouse.orderpoint'].search_count([('product_id', '=', record.x_unit_sale_product.product_variant_id.id)], limit=1)
-    bom_vals = {
-        'product_qty': quant,
-        'uom_id': record.x_unit_sale_product.uom_id.id,
-        'type': 'normal',
-        'x_parent_product': record.id,
-        'x_auto_production': True,
-        'company_id': False,
-        'bom_line_ids': [(5, 0), (0, 0, {'product_id': record.product_variant_id.id, 'product_qty': 1})],
-    }
-    if record.x_deposit_product:
-        bom_vals['byproduct_ids'] = [(5, 0), (0, 0, {'product_id': record.x_deposit_product.product_variant_id.id, 'product_qty': 1.0})]
-
-    if existing_bom: existing_bom.write(bom_vals)
-    else:
-        bom_vals['product_tmpl_id'] = record.x_unit_sale_product.id
-        env['mrp.bom'].create(bom_vals)
-    if not reordering_rule:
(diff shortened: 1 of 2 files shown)


deposit_management/data/ir_actions_server.xml
```diff
@@ -6,44 +6,53 @@ if (quant := record.x_quantity_by_deposit_product) == 1 and record.x_unit_sale_p
...
-                'route_id': env.ref('mrp.route_warehouse0_manufacture').id,
+                'route_id': manufacture_route_id,
                 'location_id': wh.lot_stock_id.id,
             })
-        env['stock.warehouse.orderpoint'].create(orderpoints_to_create)
-else:
-    existing_bom = env['mrp.bom'].search([('x_parent_product', '=',record.id)], limit=1)
-    reordering_rule = env['stock.warehouse.orderpoint'].search([('x_parent_product', '=', record.id)], limit=1)
-    if existing_bom: existing_bom.unlink()
-    if reordering_rule: reordering_rule.unlink()
```
vava-odoo, line 46: why don't you handle the unlink part?

hotel/demo/website/server_actions.xml
vava-odoo: this commit should not be here 😄

vava-odoo (commented):
Looks overall good to me, just two small comments. Thanks!

deposit_management/data/ir_actions_server.xml
jowes-odoo: In the case that the BoM/orderpoint records have been used, attempting to delete them would cause an error.

vava-odoo: Oh. Shouldn't we archive them? Or it wouldn't hurt to leave them be?

vava-odoo: And then, if there is no else, I wonder if we shouldn't add [('x_unit_sale_product', '!=', False)] as a filter_domain of the base automation, and remove this part from all conditions in the  server action

jowes-odoo: I thought about archiving, but since it doesn't hurt to keep them, I left it alone.

Moving the x_unit_sale_product check to the filter_domain makes a lot of sense. I'll make that change.

deposit_management/data/ir_actions_server.xml
```diff
@@ -5,45 +5,53 @@
 if (quant := record.x_quantity_by_deposit_product) == 1 and record.x_unit_sale_product.id:
     raise UserError("Unit sale product should be undefined if this product can not be sold in a smaller unit.")
 if quant <= 0 and record.x_unit_sale_product: raise UserError("The auto-BOM should contain a positive amount of units.")
```
vava-odoo:
```suggestion
if (quant := record.x_quantity_by_deposit_product) == 1:
   raise UserError("Unit sale product should be undefined if this product can not be sold in a smaller unit.")
if quant <= 0: raise UserError("The auto-BOM should contain a positive amount of units.")
```

deposit_management/data/ir_actions_server.xml
vava-odoo: Thanks.
Not sure about "doesn't hurt" part.
What if you configure a product, create a BoM and a RR. Then you modify the product, and remove unit product, I fear the BoM and RR still are active after a sale

vava-odoo (approved):
Just need to think about what to do with existing BoM and RR in case we remove the unit product, but as you mentioned, unlink is not an option if it has already been used, so we need to investigate on the issue.
@robodoo r+

vava-odoo:
@robodoo override=ci/industry

PR #2411, opened 2026-06-25
[FIX] industry_real_estate: fix the website which is not editable

dhrs-odoo:
In this PR we are making the website editable same as real estate industry

OPW-6306947


vava-odoo (approved):
@robodoo r+

PR #2422, opened 2026-06-26
[FIX] construction_developer: optimize work item automation triggers

chga-odoo:
Before this PR:
- Updating the price or cost of a non-template work item updated the related product's price/cost, which was not the intended behavior.
- Updating the price or cost of a work item linked to a sale order line did not update the corresponding sale order line values.

After this PR:
- Updating the price or cost of a non-template work item now correctly updates the linked sale order line's price and cost.
- The related product's price and cost are updated only when the corresponding template work item is modified.
- Add Tast cases to prevent this type of errors.

Task: 6293475

### DIFF
--- a/construction_developer/data/base_automation.xml
+++ b/construction_developer/data/base_automation.xml
@@ -54,6 +54,7 @@
         <field name="trigger">on_create_or_write</field>
         <field name="filter_domain">[("x_is_template", "=", False)]</field>
         <field name="name">Construction: Work Item - On save of non template</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('x_unit_cost_field_x_work_item'), ref('x_unit_price_field_x_work_item'), ref('x_unit_margin_field_x_work_item')])]"/>
     </record>
     <record id="base_automation_work_item_on_write_unit_cost_price_change" model="base.automation">
         <field name="model_id" ref="x_work_item_model" />
@@ -61,6 +62,7 @@
             eval="[(6, 0, [ref('server_action_update_product_cost'), ref('server_action_update_product_price')])]" />
         <field name="trigger">on_create_or_write</field>
         <field name="name">Construction: Work Item - On save and unit cost/price change</field>
+        <field name="filter_domain">[("x_is_template", "!=", False)]</field>
         <field name="trigger_field_ids" eval="[(6, 0, [ref('x_unit_cost_field_x_work_item'), ref('x_unit_price_field_x_work_item')])]"/>
     </record>
     <record id="base_automation_product_product_on_write_unit_cost_price_change" model="base.automation">
--- a/tests/test_construction_developer/tests/test_action_server.py
+++ b/tests/test_construction_developer/tests/test_action_server.py
@@ -68,14 +68,23 @@ class ActionServerTestCase(TransactionCase):
         self.assertEqual(self.test_template_work_item.x_unit_cost, 50.0, "The cost of work item is not computed correctly: it should be the sum of the product of components costs by quantity")
         self.assertEqual(self.test_template_work_item.x_unit_price, 100.0, "The price of work item is not computed correctly: it should be the sum of the product of components prices by quantity")
         self.assertEqual(self.test_template_work_item.x_unit_margin, 1.0, "The margin of work item is not computed correctly: it should be equal to 1.0 at this point")
+        self.assertEqual(self.test_product.standard_price, 50.0, "The product cost should be updated when the work item cost is updated")
+        self.assertEqual(self.test_product.lst_price, 100.0, "The product price should be updated when the work item price is updated")
+
         test_template_work_item_lines[0].x_unit_cost = 20
         self.assertEqual(self.test_template_work_item.x_unit_cost, 100.0, "The cost of work item is not recomputed correctly when the cost of a component changes")
         self.assertEqual(self.test_template_work_item.x_unit_price, 100.0, "The price of work item changes when the cost of a component changes but it should not happen")
         self.assertEqual(self.test_template_work_item.x_unit_margin, 0.0, "The margin of work item is not correct after the cost of a component changes: it should be recomputed to represent the new value")
+        self.assertEqual(self.test_product.standard_price, 100.0, "The product cost should be updated when the work item cost is updated")
+        self.assertEqual(self.test_product.lst_price, 100.0, "The product price should be updated when the work item price is updated")
+
         test_template_work_item_lines[1].x_unit_price = 240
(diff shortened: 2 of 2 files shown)


tests/test_construction_developer/tests/test_action_server.py
```diff
@@ -85,3 +85,39 @@ def test_wi_prices_correctly_computed(self):
...
+        self.test_template_work_item.x_is_margin_fixed = True
+        self.test_template_work_item.x_unit_price = 200
+        self.assertEqual(self.test_product.lst_price, 200, "The product price should be updated when the work item price is updated")
+        self.test_template_work_item.x_unit_price = 300
+        self.assertEqual(self.test_product.lst_price, 300, "The product price should be updated when the work item price is updated")
+
+        self.test_template_work_item.x_unit_cost = 100
+        self.assertEqual(self.test_product.standard_price, 100, "The product price should be updated when the work item cost is updated")
+        self.test_template_work_item.x_unit_cost = 150
+        self.assertEqual(self.test_product.standard_price, 150, "The product price should be updated when the work item cost is updated")
```
frva-odoo: While this works as a test, I find it weird as the x_unit_cost of a work item is never supposed to be changed like this (it is a computed readonly field).
I think it would be better to be close to what actually happens and instead put an x_work_item_line in the work item with a defined cost, and test if it is represented in the work item then in the product

tests/test_construction_developer/tests/test_action_server.py
```diff
@@ -85,3 +85,39 @@ def test_wi_prices_correctly_computed(self):
         self.assertEqual(self.test_template_work_item.x_unit_margin, 0.5, "When margin is fixed, the margin of work item should stay the same when the cost of a component changes")
         self.assertEqual(self.test_template_work_item.x_unit_cost, 90.0, "When margin is fixed, the cost of work item should be recomputed when the cost of a component changes")
         self.assertEqual(self.test_template_work_item.x_unit_price, 135.0, "When margin is fixed, the price of work item should be recomputed when the cost of a component changes")
+
+    def test_product_price_update_when_work_item_template_price_updated(self):
+        self.test_template_work_item.x_is_margin_fixed = True
+        self.test_template_work_item.x_unit_price = 200
+        self.assertEqual(self.test_product.lst_price, 200, "The product price should be updated when the work item price is updated")
+        self.test_template_work_item.x_unit_price = 300
+        self.assertEqual(self.test_product.lst_price, 300, "The product price should be updated when the work item price is updated")
```
frva-odoo: Same here, you only test if you directly set it, not when it is updated because of:
- margin fixed and margin changes -> price should change everywhere
- margin not fixed and work item line price change -> price should change everywhere

frva-odoo (commented):
Good job, two comments to improve/cover more test cases. Otherwise seems like a good fix 🚀

tests/test_construction_developer/tests/test_action_server.py
chga-odoo: Yes, I was aware of the flow. I took the reference from the method above, where `x_unit_price` is assigned directly, so I thought it was acceptable to follow the same approach. 😊 
Thanks for pointing it out—I’ll adapt accordingly.

pgu-odoo:

tests/test_construction_developer/tests/test_action_server.py
```diff
@@ -68,14 +68,23 @@ def test_wi_prices_correctly_computed(self):
         self.assertEqual(self.test_template_work_item.x_unit_cost, 50.0, "The cost of work item is not computed correctly: it should be the sum of the product of components costs by quantity")
         self.assertEqual(self.test_template_work_item.x_unit_price, 100.0, "The price of work item is not computed correctly: it should be the sum of the product of components prices by quantity")
         self.assertEqual(self.test_template_work_item.x_unit_margin, 1.0, "The margin of work item is not computed correctly: it should be equal to 1.0 at this point")
+        self.assertEqual(self.test_product.standard_price, 50.0, "The product price should be updated when the work item cost is updated")
```
vava-odoo: several times the same issue
```suggestion
        self.assertEqual(self.test_product.standard_price, 50.0, "The product cost should be updated when the work item cost is updated")
```

tests/test_construction_developer/tests/test_action_server.py
```diff
@@ -85,3 +94,29 @@ def test_wi_prices_correctly_computed(self):
...
+        })
+        new_work_item.write({'x_sale_order_line_id': test_sale_order_line_1.id})
+
+        self.assertEqual(test_sale_order_line_1.price_unit, new_work_item.x_unit_price, "The sale order line price should be equal to the work item price when the sale order line is created")
+        new_work_item.x_unit_price = 300
+        self.assertEqual(test_sale_order_line_1.price_unit, new_work_item.x_unit_price, "The sale order line price should be updated when the work item price is updated")
+
+        self.assertEqual(test_sale_order_line_1.purchase_price, new_work_item.x_unit_cost, "The sale order line cost should be equal to the work item cost when the sale order line is created")
+        new_work_item.x_unit_cost = 150
+        self.assertEqual(test_sale_order_line_1.purchase_price, new_work_item.x_unit_cost, "The sale order line cost should be equal to the work item cost when the sale order line is created")
```
vava-odoo:
```suggestion
        self.assertEqual(test_sale_order_line_1.purchase_price, new_work_item.x_unit_cost, "The sale order line cost should be equal to the work item cost when the sale order line is updated")
```

construction_developer/data/base_automation.xml
```diff
@@ -54,13 +54,15 @@
...
         <field name="name">Construction: Work Item - On save of non template</field>
+        <field name="trigger_field_ids" eval="[(6, 0, [ref('x_unit_cost_field_x_work_item'), ref('x_unit_price_field_x_work_item'), ref('x_unit_margin_field_x_work_item')])]"/>
     </record>
     <record id="base_automation_work_item_on_write_unit_cost_price_change" model="base.automation">
         <field name="model_id" ref="x_work_item_model" />
         <field name="action_server_ids"
             eval="[(6, 0, [ref('server_action_update_product_cost'), ref('server_action_update_product_price')])]" />
         <field name="trigger">on_create_or_write</field>
         <field name="name">Construction: Work Item - On save and unit cost/price change</field>
+        <field name="filter_domain">[("x_is_template", "!=", False)]</field>
```
vava-odoo, line 65: You're not testing this, right?

chga-odoo: No, But now it is :smile:

vava-odoo (approved):

vava-odoo:
@robodoo r+

PR #2424, opened 2026-06-26
[IMP] *: adapt category of technical modules

vava-odoo:
This commit adapts the category of technical modules such that it can be retrieved and installed from the database directly.

task-6341260


vava-odoo:
@robodoo r+

PR #2426, opened 2026-06-29
[FIX] booking_engine: fix navigation to the guest view

dhrs-odoo:
This PR removes the **many2one_avatar** widget, which prevented users from navigating to the guest view and modifying guest data.

Task-6273305


pgu-odoo:

vava-odoo (approved):
@robodoo r+

PR #2427, opened 2026-06-29
[FIX] hvac_services: enable serial numbers on invoices

dhrs-odoo:
This PR enables the **Display Lots & Serial Numbers on Invoices** setting to ensure serial numbers are displayed on customer invoices.

Task-6329630

### DIFF
--- a/hvac_services/data/qweb_view.xml
+++ b/hvac_services/data/qweb_view.xml
@@ -124,4 +124,25 @@
             </t>
         </field>
     </record>
+    <template id="report_invoice_document_serials" inherit_id="account.report_invoice_document">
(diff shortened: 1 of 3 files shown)


hvac_services/data/qweb_view.xml
```diff
@@ -124,4 +124,25 @@
             </t>
         </field>
     </record>
+    <template id="report_invoice_document_serials" inherit_id="account.report_invoice_document">
+        <xpath expr="//t[@t-set='display_taxes']" position="after">
```
vava-odoo, line 128: Why is there no check on the setting?

vava-odoo (changes requested):

vava-odoo (approved):
@robodoo r+

PR #2434, opened 2026-06-30
[FIX] .*: restore and align website headers

dhrs-odoo:
Before this PR, the website headers were commented out because they did not behave as expected.

This PR fixes the header implementation and aligns it with the website design.

**Task-6345228**

### DIFF
--- a/concert_halls/demo/website_theme_apply.xml
+++ b/concert_halls/demo/website_theme_apply.xml
@@ -67,17 +67,15 @@
     <value eval="{'active': False}"/>
   </function>
 
-  <!-- Because the header template wasn’t working correctly, I have commented it out -->
-
-  <!-- <function model="ir.ui.view" name="write">
-    <value model="ir.ui.view" eval="obj().search([('key', '=', 'website.template_header_default')], limit=1).id"/>
+  <function model="ir.ui.view" name="write">
+    <value model="ir.ui.view" eval="obj().search([('key', '=', 'website.template_header_default')]).ids"/>
     <value eval="{'active': False}"/>
   </function>
 
   <function model="ir.ui.view" name="write">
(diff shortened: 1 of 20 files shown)


concert_halls/demo/website_theme_apply.xml
```diff
@@ -67,17 +67,15 @@
...
-  <!-- <function model="ir.ui.view" name="write">
-    <value model="ir.ui.view" eval="obj().search([('key', '=', 'website.template_header_default')], limit=1).id"/>
+  <function model="ir.ui.view" name="write">
+    <value model="ir.ui.view" eval="obj().search([('key', '=', 'website.template_header_default')]).ids"/>
     <value eval="{'active': False}"/>
   </function>
 
   <function model="ir.ui.view" name="write">
-    <value model="ir.ui.view" eval="obj().search([('key', '=', 'website.template_header_vertical')], limit=1).id"/>
+    <value model="ir.ui.view" eval="obj().with_context(active_test=False).search([('key', '=', 'website.template_header_vertical')], limit=1).id"/>
```
pgu-odoo, line 76:
```suggestion
    <value model="ir.ui.view" eval="obj().with_context(active_test=False).search([('key', '=', 'website.template_header_vertical')]).ids"/>
```
Can we remove limit=1 just an improvement? and pass the ids to write method?
as we are passing key to search method, mostly only one record will be there and if many, all will be active true.

or even why we are not using ref?

dhrs-odoo: Here, we only have a single `website.template_header_vertical` record, and it is inactive. If we use `ref` directly, the record will not be found because inactive records are excluded by default. Therefore, we need to search with `active_test=False`.

Since we only want to activate the intended view, we also use `limit=1` to ensure that no other view with the same key is accidentally activated.

pgu-odoo:

vava-odoo (approved):
@robodoo r+

PR #2471, opened 2026-07-02
[FIX] 3pl_logistic_company: changed barcode value

bhna-odoo:
Currently, an error occurs when user tries to install 3pl_logistic_company.

Steps to replicate:
- Install `stock` with demo.
- Install `3pl_logistic_company`.

Error:
 ```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "stock_package_type_barcode_uniq" DETAIL:  Key (barcode)=(PAL) already exists.
```

Cause:
- [1] and [2] both the records have the same barcode value so the [unique] constraint causes this error.

Solution:
- Changed the barcode value.

[1]: https://github.com/odoo/industry/blob/58e530739f28dc79d97f57cadb39bc778a32ca12/3pl_logistic_company/data/stock_package_type.xml#L7

[2]: https://github.com/odoo/odoo/blob/2d54db3ac0b6d807e580315e2633f3e2b10a700c/addons/stock/data/stock_demo.xml#L24

[unique]: https://github.com/odoo/odoo/blob/2d54db3ac0b6d807e580315e2633f3e2b10a700c/addons/stock/models/stock_package_type.py#L41-L44

No ID

### DIFF
--- a/3pl_logistic_company/__manifest__.py
+++ b/3pl_logistic_company/__manifest__.py
@@ -1,7 +1,7 @@
 {
     'name': '3PL Logistic Company',
     'author': 'Odoo S.A.',
-    'version': '1.5',
+    'version': '1.6',
     'category': 'Services',
     'depends': [
         'base_industry_data',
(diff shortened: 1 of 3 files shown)


maan-odoo (approved):
LGTM

vava-odoo (approved):
Thanks @bhna-odoo 
We don't usually fix compatibility issues with standard demo data, because industry modules come with their own set of demo data.
But I bet this case is simple enough to fix. 
Could you just update the name of all stock.package records in demo folder to reflect this change?

vava-odoo:
@robodoo r+

PR #2491, opened 2026-07-06
[FIX] agriculture_store: modified the description for the appstore

dhrs-odoo:
In this commit we have fixed the apps store description which was appearing distorted with the earlier code by commit -
https://github.com/odoo/industry/commit/3d31f216102d9b2aef4c44976a037ae74d361bf3

Task-6347835


vava-odoo (approved):
@robodoo r+

PR #2497, opened 2026-07-07
[FIX] booking_engine: Preserve planning slot view for inherited xpaths

amas-odoo:
The `booking_engine` replaced the `sale` container in the planning view, removing the `sale_line_id_div` element. As a result, modules such as `planning_field_service_sale_timesheet`, which inherit the view and extend `sale_line_id_div`, failed to install with an xpath error.

Instead of replacing the sale container, mark it as invisible. This preserves the original view structure, allowing inherited xpaths to continue working while still hiding the sale information in the booking engine.

Task ID: 6368237

### DIFF
--- a/booking_engine/__manifest__.py
+++ b/booking_engine/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Booking',
-    'version': '1.22',
+    'version': '1.23',
     'category': 'Hospitality',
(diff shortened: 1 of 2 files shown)


dhrs-odoo (approved):
LGTM 🚀

vava-odoo:
@robodoo r+

PR #2505, opened 2026-07-08
[IMP] booking_engine: remove useless stock dependency

vava-odoo:
To prevent double booking from ecommerce, this module relied on stock warning saying that the product was not available. This warning recently moved from website_sale_stock_renting to website_sale_renting. This was the only reason stock was in the dependency, so it can be removed.


abdrahmanrashed (approved):

vava-odoo:
@robodoo r+

PR #2514, opened 2026-07-09
[FIX] *: fix remaining upgrade issues to 19.4

amas-odoo:
This commit fixes a few issues in the code and tests:
- ensure well-defined domains in booking_engine and textile_manufacturing;
- add new step to create order in condominium tour;
- replace orm_service by orm_plugin in js patch of construction;
- add missing file in manifest for hotel;
- add missing dependency in vineyard.

PR #2529, opened 2026-07-10
[FIX] textile_manufacturing: Fix domains

amas-odoo:
This commit ensure well-defined domains in textile_manufacturing

(backport from https://github.com/odoo/industry/pull/2514)


vava-odoo:
@robodoo r+ up to saas-19.3

PR #2544, opened 2026-07-13
[FIX] industry_real_estate: prevent out of memory error on attachment related fields

m-gad1:
Issue:
Since v19.0, an updated access check on ir.attachment records causes an out-of-memory error when loading models with related fields to attachments, specifically during the fields_get() call.
Standard Python-defined fields bypass this with `bypass_search_access=True` (in #217277).
Industry modules define fields in XML, where this can't be applied.

Fix:
Add a domain to the fields definition to restrict the query result.

opw-6348453

### DIFF
--- a/industry_real_estate/data/ir_model_fields.xml
+++ b/industry_real_estate/data/ir_model_fields.xml
@@ -277,6 +277,7 @@ for property in self:
         <field name="name">x_property_attachment_image_ids</field>
         <field name="field_description">Property Images</field>
         <field name="model_id" ref="analytic.model_account_analytic_account"/>
+        <field name="domain">[('res_model', '=', 'account.analytic.account')]</field>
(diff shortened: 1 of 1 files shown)


vava-odoo (commented):
This is not a fix, but a patch hiding the real problem.
@kmagusiak do you know about this issue (memory usage going crazy with ir.attachments)?

kmagusiak:
> This is not a fix, but a patch hiding the real problem. @kmagusiak do you know about this issue (memory usage going crazy with ir.attachments)?

Yes. If you reference an attachment or a message, we need to check the access and if you try to search on that comodel without a res_model, the only thing we can do is to filter everything in memory.

vava-odoo:
Thanks!

@robodoo r+

PR #2546, opened 2026-07-13
[FIX] *: define trigger fields if domain on automation

vava-odoo:
This commit is a partial backport of a fix[^1] in saas-19.3. This is backported to improve the triggering on automations, improve efficiency of server action or fix invalid domain for automations.

[^1]: https://github.com/odoo/industry/commit/9328320b8761398309ea2a064e1d4442a17c60d8

### DIFF
--- a/art_craft/__manifest__.py
+++ b/art_craft/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Arts & Crafts Store',
-    'version': '1.1',
+    'version': '1.2',
     'category': 'Retail',
     'depends': [
         'hr_hourly_cost',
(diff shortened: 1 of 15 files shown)


frva-odoo (approved):
Shouldn't we backport the test too ? 🤔 
Otherwise all good, let's not forget in 19.3 the bump for construction_developer, hair_salon, handyman and headhunter; with these we should have covered all modules

vava-odoo:
But the test is in 19.3 because the behaviour changed there. Adding an empty trigger field set is useless in older version. Right?

frva-odoo:
> But the test is in 19.3 because the behaviour changed there. Adding an empty trigger field set is useless in older version. Right?

You're right, that is not happening version 19.2 or before. So yes I'm wrong sorry, backporting the test makes no sense

vava-odoo:
@robodoo rebase-ff

vava-odoo:
@robodoo r+

PR #2553, opened 2026-07-14
[FIX] bar_industry: fix appstore description and video reference

dhrs-odoo:
In this PR we have updated the appstore description with the standard layout and also added the video reference for the industry flow

Task-6352086


vava-odoo:
@robodoo r+

PR #2577, opened 2026-07-15
[FIX] headhunter: fix CV publicly available

frva-odoo:
In the headhunter industry, to able the recruiters to see the CVs, we previously set the documents as public.

Security wise, this is not the best choice and so this commit changes that, by keeping the documents private and using a token in the cv portal page instead.

Task-6389177


vava-odoo:
moved to https://github.com/odoo/industry/pull/2579

PR #2579, opened 2026-07-15
[FIX] headhunter: fix demo data and modify shared cv access

frva-odoo:
This commit is fixing the demo "cv applicant lines" by adding
the CV attachments to the records. This is also modifying the way
we handle the access of said attachments by using tokens in our links.

Task-6389177


vava-odoo (approved):
@robodoo r+

PR #2580, opened 2026-07-15
[FIX] food_trucks: do not raise if pos.session does not exist

vava-odoo:
Before this commit, the steps to close and re-open standard pos.session in order to edit the pos.config failed without demo data because the session does not exist.

This commit adds `raise_if_not_found=False` to the ref in order to skip the closing of the pos.session if this one does not exist.

### DIFF
--- a/food_trucks/data/pos_config.xml
+++ b/food_trucks/data/pos_config.xml
@@ -5,7 +5,7 @@
   </function>
   <!-- temporarily close the session to remove the floor plans -->
   <function model="pos.session" name="write">
-      <value model="pos.session" eval="ref('pos_restaurant.pos_open_session_2')"/>
+      <value model="pos.session" eval="ref('pos_restaurant.pos_open_session_2', raise_if_not_found=False)"/>
       <value eval="{'state': 'closed'}"/>
   </function>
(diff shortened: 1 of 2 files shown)


food_trucks/data/pos_config.xml
```diff
@@ -5,15 +5,15 @@
   </function>
   <!-- temporarily close the session to remove the floor plans -->
   <function model="pos.session" name="write">
-      <value model="pos.session" eval="ref('pos_restaurant.pos_open_session_2')"/>
+      <value model="pos.session" eval="ref('pos_restaurant.pos_open_session_2', raise_if_not_found=False)"/>
```
trcazier, line 8: Will `write` work with `None` as argument? Other than that, LGTM

trcazier (approved):

food_trucks/data/pos_config.xml
vava-odoo: Yes. `write` starts with 
```py
if not self:
    return True
```
I ran the single industry build and it works in both mode (demo / no-demo), where it failed before in no-demo.

vava-odoo: thanks for the review :pray:

vava-odoo:
@robodoo r+ squash

PR #2585, opened 2026-07-15
[FIX] campsite: fix stock dependency error

frva-odoo:
The stock dependency was previously acquired through pos_restaurant, but that is not the case anymore.
That's why the module stopped working with demo datas. Instead of adding stock in the dependencies, we decided to remove it by cleaning the records related to stock and use ministock to display inventory instead.

Task-6384179


vava-odoo (approved):

vava-odoo:
@robodoo r+

PR #2586, opened 2026-07-15
[ADD] property_listing: Added sub module Property Listing

amas-odoo:
- Extract the common functionality shared by Real Estate and Property Management
into a new technical module Property Listing.
- Make the Real Estate industry depend on the new module.
- Move shared website shop configuration and customizations to Property Listing.

Task ID: 5500297


pgu-odoo:
@amas-odoo  :clap:

property_listing/data/website_view.xml
```diff
@@ -0,0 +1,181 @@
...
+                        </section>
+                    </div>
+                </t>
+            </t>
+        </field>
+    </record>
+    <function model="ir.model.fields" name="formbuilder_whitelist">
+        <value eval="'crm.lead'"/>
+        <value eval="['x_property_id']"/>
+    </function>
```
vava-odoo: You can do it at the field level setting `website_form_blacklisted` to True
```suggestion
```

property_listing/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+{
+    'name': 'Property Listing',
+    'category': 'Services',
+    'author': 'Odoo S.A.',
+    'depends': [
+        'crm_enterprise',
+        'web_studio',
+        'website_sale',
```
vava-odoo, line 9: since we create leads from the form, we need website_crm dependency

industry_real_estate/__manifest__.py
```diff
@@ -1,24 +1,19 @@
 {
     'name': 'Property Management',
-    'version': '2.3',
+    'version': '2.4',
```
vava-odoo: 3.0

real_estate/data/ir_model_fields.xml
```diff
@@ -592,7 +576,7 @@ for lead in self:   lead['x_sign_request_count'] = lead.env['sign.request'].sear
         <field name="relation">product.template</field>
         <field name="depends">x_crm_lead_id</field>
         <field name="compute"><![CDATA[ 
-for request in self:    request['x_property_id'] = request.x_crm_lead_id.x_interested_in_id if request.x_crm_lead_id and request.x_crm_lead_id.x_interested_in_id else False
+for request in self:    request['x_property_id'] = request.x_crm_lead_id.x_property_id if request.x_crm_lead_id and request.x_crm_lead_id.x_property_id else False
```
vava-odoo:
```suggestion
for request in self:    request['x_property_id'] = request.x_crm_lead_id.x_property_id if request.x_crm_lead_id.x_property_id else False
```

real_estate/__manifest__.py
vava-odoo:
```suggestion
    'version': '2.0',
```

real_estate/__manifest__.py
```diff
@@ -5,20 +5,15 @@
...
-        'base_automation',
         'base_industry_data',
-        'crm_enterprise',
         'documents',
         'hr',
         'knowledge',
         'project_enterprise',
+        'property_listing',
         'sale_commission',
-        'sale_crm',
```
vava-odoo, line 15: this probably is also required in property_listing

vava-odoo (commented):
This looks almost fine.
I'd do it in master, though.

My major concern is the fact that we merge two property fields on crm.lead into a single one. I thought they had a very different purpose, so I'm afraid we miss part of the flow
@Pierre-savp @Jean-boje

Pierre-savp:
> This looks almost fine. I'd do it in master, though.
> 
> My major concern is the fact that we merge two property fields on crm.lead into a single one. I thought they had a very different purpose, so I'm afraid we miss part of the flow @Pierre-savp @Jean-boje

@vava-odoo 
the x_property_id on crm.lead is coming from website and leveraged in real_estate to seggregate leads into pipelines per product.template (properties)
goal is to have the exact same behavior in industry_real_estate as well since in a similar fashion, when they have a property to rent, a dedicated pipeline is great
this is part of the "more in common" second step to leverage all the value from real_estate : https://www.odoo.com/odoo/my-tasks/6394306

which other use case are you referring to?

vava-odoo:
@Pierre-savp in real_estate, there two major flows: sell and buy. Having two fields was convenient to easily separate both cases in domain, actions, views, ...
I didn't really review functionally the business flows, I just feel like we'll miss something for sure

property_listing/__manifest__.py
amas-odoo: Do we really need to add this dependency? As far as I can see, website_crm is installed, and I'm able to create leads from the website without explicitly adding it as a dependency. Please refer to the attached screenshot (SS).

image
image

real_estate/__manifest__.py
amas-odoo: This also installed.

image

amas-odoo:
> @Pierre-savp in real_estate, there two major flows: sell and buy. Having two fields was convenient to easily separate both cases in domain, actions, views, ... I didn't really review functionally the business flows, I just feel like we'll miss something for sure

@vava-odoo I think I initially missed the purpose of having two separate fields. However, maintaining two Many2one fields doesn't seem like the best approach. We could instead have a single Many2one field for the property and add a separate selection field (e.g., Sell or Buy) to distinguish the lead's intent. In my opinion, that would be a cleaner and more maintainable solution.

If you and @Pierre-savp agree with this approach, I'll make the changes accordingly. Or if not, I will revert.

https://github.com/odoo/industry/blob/d9369bec99e2dabd4284d0e1230551fc342b9a1b/real_estate/data/ir_model_fields.xml#L122
https://github.com/odoo/industry/blob/d9369bec99e2dabd4284d0e1230551fc342b9a1b/real_estate/data/ir_model_fields.xml#L186

Pierre-savp:
> > @Pierre-savp in real_estate, there two major flows: sell and buy. Having two fields was convenient to easily separate both cases in domain, actions, views, ... I didn't really review functionally the business flows, I just feel like we'll miss something for sure
> 
> @vava-odoo I think I initially missed the purpose of having two separate fields. However, maintaining two Many2one fields doesn't seem like the best approach. We could instead have a single Many2one field for the property and add a separate selection field (e.g., Sell or Buy) to distinguish the lead's intent. In my opinion, that would be a cleaner and more maintainable solution.
> 
> If you and @Pierre-savp agree with this approach, I'll make the changes accordingly. Or if not, I will revert.
> 
> https://github.com/odoo/industry/blob/d9369bec99e2dabd4284d0e1230551fc342b9a1b/real_estate/data/ir_model_fields.xml#L122
> 
> 
> https://github.com/odoo/industry/blob/d9369bec99e2dabd4284d0e1230551fc342b9a1b/real_estate/data/ir_model_fields.xml#L186

fine for me, even though i haven't captured any "collision" during tests but may be preferable to distinguish at this stage those fields
let me know when done i'll make another round of testing

property_listing/__manifest__.py
vava-odoo: it is auto installed with website and crm. But I wonder if we shouldn't enforce the dependency. @Pierre-savp maybe?

real_estate/__manifest__.py
vava-odoo: I didn't fully check, but if you base a view or an action (or else) on this module, then it needs to be in the dependency tree

property_listing/__manifest__.py
Pierre-savp: makes sense at least to be future proof in case the dependency path changes i assume

vava-odoo:
@Pierre-savp I agree with @amas-odoo about the fact that having 2 m2o is not needed. But I still think we should be cautious, so I checked a bit more in depth, and here is where the flow may be impacted:
- Automation: On new lead, there is an automation (On new lead) to auto-assign properties based on the email the user sent (if there is any reference in it). This will run on all leaves where a property is not selected any time a message is received on the lead.I don't think this creates any issue here.
- Window actions: I feel like the distinction is not entirely clear now between Property > Buyers and CRM (sellers). I see some leads in both views, where this shouldn't be the case.
- Smart buttons: since this is the same field, the seller is included in the number of interested leads on properties.
- Reporting: graph and pivot views don't distinguish sell and buy cases.

As you can see, no big deal in my opinion, but I wonder if an extra step is not required here to have it perfectly clean. Let us know

Pierre-savp:
@vava-odoo
thanks for the use case clarifications, was challenging to capture all side effects...

> Automation: On new lead, there is an automation (On new lead) to auto-assign properties based on the email the user sent (if there is any reference in it). This will run on all leaves where a property is not selected any time a message is received on the lead.I don't think this creates any issue here.

Not an issue

> Window actions: I feel like the distinction is not entirely clear now between Property > Buyers and CRM (sellers). I see some leads in both views, where this shouldn't be the case.

That is not suitable

> Smart buttons: since this is the same field, the seller is included in the number of interested leads on properties.

This neither

> Reporting: graph and pivot views don't distinguish sell and buy cases.

Same here

**All in all there are indeed 2 usage of m2o property on lead so i'd keep them separated using relavant naming by perimeter**
proposal: 
* interested_in for buyers/renters
* owner_of for owners

sorry for the back and forse @amas-odoo

industry_real_estate/data/website_theme_apply.xml
```diff
@@ -4,4 +4,8 @@
         <value model="ir.ui.view" eval="obj().env['website'].with_context(website_id=obj().env.ref('base.default_website').id).viewref('website.contactus').id"/>
         <value model="ir.ui.view" eval="{'arch': obj().env.ref('industry_real_estate.contactus').arch}"/>
     </function>
+    <function model="ir.ui.view" name="write">
+        <value model="ir.ui.view" eval="obj().env['website'].with_context(website_id=obj().env.ref('base.default_website').id).viewref('website.contactus').id"/>
```
vava-odoo, line 4: can't you do this in the write just above?

real_estate/data/website_view.xml
```diff
@@ -288,115 +288,6 @@
...
-                </div>
-              </section>
-          </div>
-        </t>
-      </t>
-    </field>
-  </record>
   <record id="ir_ui_view_3022" model="ir.ui.view">
     <field name="inherit_id" ref="website_sale.header_cart_link"/>
     <field name="key">website_sale.header_hide_empty_cart_link</field>
```
vava-odoo: do we still need this?

vava-odoo (commented):
very minor comments
@Pierre-savp did you do a last functionnal review here?
Thanks!

Pierre-savp:
@amas-odoo @vava-odoo 
checked based on yesterday last push and works for me
only drawback is the duplication of properties smart button on the contact form when having both industries...
we'll work on this later and maybe we can go for a merge of product and x_property models to rationalize... we'll see later ;)
GOOD FOR NOW ;)

vava-odoo (approved):
@robodoo r+ squash

PR #2610, opened 2026-07-16
[FIX] micro_brewery: Fix demo data for BoMs

jowes-odoo:
mrp_bom_17 and mrp_bom_18 are created for two products that are using
the auto-bom feature. This results in duplicate BoMs.

To fix, we can remove the BoM definition and let the BoMs be
created by the automation.

opw-6395817


vava-odoo (approved):
@robodoo r+

PR #2643, opened 2026-07-24
[FIX] tests: remove knowledge article publication tes

dhrs-odoo:
This commit removes the `_check_knowledge_article_is_published` test.

The test verified that the `is_published` field exists on knowledge articles and displayed a warning when it was present. Since this check is no longer required, the test has been removed.


vava-odoo:
@robodoo r+

PR #2644, opened 2026-07-24
[IMP] bookstore: improve data and remove automatic purchase order creation

dhgoy-odoo:
- Add supplier information for demo products to support replenishment
- Rename the replenishment filter in the Knowledge article
- Update the "Encode new release" workflow
- Remove outdated reporting instructions
- Simplify the "Update supplier info on incoming products" workflow by 
removing automatic purchase order creation

TaskId-6391259

### DIFF
--- a/bookstore/data/ir_actions_server.xml
+++ b/bookstore/data/ir_actions_server.xml
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo>
     <record id="ir_act_server_set_condominiums_on_partner" model="ir.actions.server">
-        <field name="name">Execute Code</field>
+        <field name="name">Create supplier info &amp; replenishment rule</field>
         <field name="state">code</field>
         <field name="base_automation_id" ref="base_automation_1"/>
         <field name="model_id" ref="stock.model_stock_picking"/>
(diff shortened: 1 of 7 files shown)


bookstore/data/ir_actions_server.xml
```diff
@@ -1,46 +1,31 @@
...
-            'delay': 0,
-            'min_qty': 1,
-        })
+        if move.product_id.seller_ids[:1].partner_id != picking.partner_id:
+            move.product_id.seller_ids.filtered(lambda s: not s.date_end).write({'date_end': datetime.datetime.today().date()})
+            env['product.supplierinfo'].create({
+                'product_id': move.product_id.id,
+                'partner_id': picking.partner_id.id,
+                'date_start': datetime.datetime.today().date(),
+                'sequence': (move.product_id.seller_ids[:1].sequence) - 1,
```
vava-odoo, line 19: why did you remove the fallback?

vava-odoo (approved):
maybe insist in the commit title/message on the major change here, the server action does not create a purchase anymore

bookstore/data/ir_actions_server.xml
dhgoy-odoo: I removed the fallback because when `sequence` becomes `0`, `(sequence or 10)` evaluates to `10`, resulting in a new sequence of `9`. For example, if an existing supplier has a sequence of `0`, the newly created supplier would get a sequence of `9`, so it would no longer be placed at the top. That's why I removed the fallback.

vava-odoo:
@robodoo r+

PR #2647, opened 2026-07-24
[IMP] hotel+: add default Booking Engine dashboard

amas-odoo:
Implement a day-to-day dashboard for accommodation industries and expose it as the default landing page of the Booking Engine app.

[IMP] booking_engine: Refine housekeeping task deadlines
- Only create housekeeping tasks for planning slots linked to a sale order line.
- Set the task deadline to the associated stay offer check-in time instead of a
fixed value, improving task visibility on the dashboard.

Task: 6205260

### DIFF
--- a/booking_engine/data/spreadsheet_dashboard.xml
+++ b/booking_engine/data/spreadsheet_dashboard.xml
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo noupdate="1">
     <record id="spreadsheet_dashboard_group_accomodation" model="spreadsheet.dashboard.group">
-        <field name="name">Accomodation</field>
+        <field name="name">Accommodation</field>
         <field name="sequence">0</field>
     </record>
     <record id="spreadsheet_dashboard_1" model="spreadsheet.dashboard">
@@ -10,4 +10,23 @@
         <field name="dashboard_group_id" ref="spreadsheet_dashboard_group_accomodation"/>
         <field name="group_ids" eval="[(6, 0, [ref('base.group_system')])]"/>
     </record>
+    <record id="spreadsheet_dashboard_2" model="spreadsheet.dashboard">
+        <field name="spreadsheet_binary_data" type="bytes" file="booking_engine/static/src/binary/spreadsheet_dashboard/Day-to-day-Operations.osheet.json"/>
+        <field name="name">Day-to-Day Operations</field>
+        <field name="dashboard_group_id" ref="spreadsheet_dashboard_group_accomodation"/>
+        <field name="group_ids" eval="[(6, 0, [ref('base.group_user')])]"/>
+    </record>
+
+    <function model="spreadsheet.dashboard" name="write">
+        <value model="spreadsheet.dashboard" eval="obj().env.ref('booking_engine.spreadsheet_dashboard_2').id"/>
+        <value model="spreadsheet.dashboard" eval="{
+            'spreadsheet_data': (
+                obj().env.ref('booking_engine.spreadsheet_dashboard_2').spreadsheet_data
+            ).replace(
+                '*project_id*', str(obj().env.ref('booking_engine.project_project_1').id)
+            ).replace(
+                '*ready_project_stage*', str(obj().env.ref('booking_engine.project_task_type_18').id)
+            )
+        }"/>
+    </function>
 </odoo>
--- a/booking_engine/static/src/binary/spreadsheet_dashboard/Day-to-day-Operations.osheet.json
+++ b/booking_engine/static/src/binary/spreadsheet_dashboard/Day-to-day-Operations.osheet.json
@@ -1,0 +1,684 @@
+{
+  "version": "19.3.10",
+  "sheets": [
+    {
+      "id": "sheet1",
+      "name": "Dashboard",
+      "colNumber": 14,
+      "rowNumber": 39,
+      "rows": {},
+      "cols": {
+        "0": {
+          "size": 29
+        },
+        "1": {
+          "size": 29
+        },
+        "2": {
+          "size": 170
+        },
+        "3": {
+          "size": 197
+        },
+        "4": {
+          "size": 20
+        },
+        "5": {
+          "size": 29
+        },
+        "6": {
+          "size": 170
+        },
+        "7": {
+          "size": 197
+        },
+        "8": {
+          "size": 20
+        },
+        "9": {
+          "size": 99
+        },
+        "10": {
+          "size": 99
+        },
+        "11": {
+          "size": 99
+        },
+        "12": {
+          "size": 99
+        },
+        "13": {
+          "size": 29
+        }
+      },
+      "merges": [],
+      "cells": {
+        "B2": "[Check In](odoo://view/{\"viewType\":\"graph\",\"action\":{\"domain\":[[\"state\",\"not in\",[\"draft\",\"cancel\",\"sent\"]]],\"context\":{\"group_by\":[\"date:month\"],\"graph_measure\":\"price_subtotal\",\"graph_mode\":\"line\",\"graph_groupbys\":[\"date:month\"]},\"modelName\":\"sale.report\",\"views\":[[false,\"graph\"],[false,\"pivot\"],[false,\"search\"]]},\"threshold\":0,\"name\":\"Sales Analysis\"})",
+        "B5": "[Orders](odoo://view/{\"viewType\":\"kanban\",\"action\":{\"xmlId\":\"booking_engine.booking_engine_rooms_order_action\",\"domain\":[\"&\",[\"x_order_involves_room\",\"=\",true],\"&\",[\"is_rental_order\",\"=\",true],\"&\",\"&\",[\"next_action_date\",\">=\",\"today\"],[\"next_action_date\",\"<\",\"today +1d\"],[\"rental_status\",\"=\",\"pickup\"]],\"context\":{\"group_by\":[]},\"modelName\":\"sale.order\",\"views\":[[false,\"kanban\"],[false,\"list\"],[false,\"form\"],[false,\"search\"]]},\"threshold\":0,\"name\":\"Orders\"})",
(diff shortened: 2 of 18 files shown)


booking_engine/data/spreadsheet_dashboard.xml
```diff
@@ -1,13 +1,33 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo noupdate="1">
-    <record id="spreadsheet_dashboard_group_accomodation" model="spreadsheet.dashboard.group">
-        <field name="name">Accomodation</field>
+    <record id="spreadsheet_dashboard_group_accommodation" model="spreadsheet.dashboard.group">
```
trcazier: Won't this create a duplicate records on upgrade if you're modifying the ID of an already existing record?

booking_engine/data/spreadsheet_dashboard.xml
```diff
@@ -1,13 +1,33 @@
...
-        <field name="dashboard_group_id" ref="spreadsheet_dashboard_group_accomodation"/>
+        <field name="dashboard_group_id" ref="spreadsheet_dashboard_group_accommodation"/>
         <field name="group_ids" eval="[(6, 0, [ref('base.group_system')])]"/>
     </record>
+    <record id="spreadsheet_dashboard_2" model="spreadsheet.dashboard">
+        <field name="spreadsheet_binary_data" type="bytes" file="booking_engine/static/src/binary/spreadsheet_dashboard/Day-to-day-Operations.osheet.json"/>
+        <field name="name">Day-to-Day Operations</field>
+        <field name="dashboard_group_id" ref="spreadsheet_dashboard_group_accommodation"/>
+        <field name="group_ids" eval="[(6, 0, [ref('base.group_user')])]"/>
+        <field name="x_is_accommodation_dashboard" eval="True"/>
```
trcazier: Can you explain why the first one has `base.group_system` and the second one has `base.group_user`? Is this important ? (I really don't know how this works)

trcazier: Also I see you changed `spreadsheet_snapshot` to `spreadsheet_binary_data`, can you explain the difference?
I see `spreadsheet_snapshot` is for a `spreadsheet.template` and `spreadsheet_binary_data` is for `spreadsheet.mixin` and I can't tell what they're used for

booking_engine/static/src/binary/spreadsheet_dashboard/Day-to-day-Operations.osheet.json
```diff
@@ -0,0 +1 @@
...,"today"],["date_deadline","<","today +1d"],["stage_id.id","!=","*ready_project_stage*"]],"filters":[],"formulaId":"7","fieldMatching":{}}},"pivotNextId":9,"customTableStyles":{},"namedRanges":{},"globalFilters":[],"lists":{"1":{"model":"sale.order","name":"Sales Order","columns":[{"name":"name","string":"Number","hidden":false},{"name":"partner_id","string":"Customer","hidden":false}],"domain":["&","&","&",["rental_start_date",">=","today"],["rental_start_date","<","today +1d"],["rental_status","=","pickup"],["order_line.product_template_id.x_accommodation_product","=","early_checkin"]],"context":{},"orderBy":[{"name":"rental_start_date","asc":false}],"fieldMatching":{}},"2":{"model":"sale.order","name":"Sales Order","columns":[{"name":"name","string":"Number"},{"name":"partner_id","string":"Customer"}],"domain":["&","&","&",["rental_return_date",">=","today"],["rental_return_date","<","today +1d"],["rental_status","=","return"],["order_line.product_template_id.x_accommodation_product","=","late_checkout"]],"context":{},"orderBy":[],"fieldMatching":{}}},"listNextId":3,"odooLinkReferences":{"222e338b-b0ce":{"type":"odooMenu","odooMenuId":"booking_engine.menu_tasks_house_keeping"}}}
```
trcazier: I think it'd be better if it was readable 😅

booking_engine/data/ir_model_fields.xml
```diff
@@ -1043,4 +1043,10 @@ for record in self:
         <field name="relation">x_so_daily_line</field>
         <field name="relation_field">x_sale_order_line_id</field>
     </record>
+    <record id="x_is_accommodation_dashboard_spreadsheet_dashboard" model="ir.model.fields">
+        <field name="name">x_is_accommodation_dashboard</field>
+        <field name="ttype">boolean</field>
+        <field name="field_description">Is accommodation dashboard</field>
+        <field name="model_id" ref="spreadsheet_dashboard.model_spreadsheet_dashboard"/>
+    </record>
```
trcazier: I get that this is probably to differentiate `spreadsheet_dashboard_1` and `2` for some reason, but is this the only way to do it? Is it not possible to use refs for example?

booking_engine/data/spreadsheet_dashboard.xml
amas-odoo: First things first:
If `base.group_system` is used, only administrators (Settings users) can access it, so the dashboard won't be visible to normal users.
Using base.group_user makes the dashboard available to all internal users, which is the intended behavior in this case.

And from my understanding:

- `spreadsheet_binary_data` contains the original spreadsheet. It only changes if the user explicitly replaces the spreadsheet, for example by importing a new XLSX, JSON, or CSV file.
- `spreadsheet_snapshot` stores the latest saved state of the spreadsheet. When users make changes, those changes update the snapshot, not `spreadsheet_binary_data` or `spreadsheet_data`

When opening a spreadsheet, Odoo first checks if a snapshot exists:

- If it does, Odoo loads the snapshot and then applies any revisions made after it.
- If no snapshot exists, Odoo loads `spreadsheet_binary_data` as the starting point.

https://github.com/odoo/enterprise/blob/03b0fa2fd0ce73788b87ff4b2fae493c2a595ad7/spreadsheet_edition/models/spreadsheet_mixin.py#L365

So creating or updating a snapshot does not modify `spreadsheet_binary_data`; the original spreadsheet remains unchanged.

One thing I'm still trying to understand is why we use `spreadsheet_snapshot` above. As far as I can see, all the dashboard implementations in the base code use `spreadsheet_binary_data`. Is there a specific reason we should use the snapshot instead of the original spreadsheet data in this case?

booking_engine/data/ir_model_fields.xml
amas-odoo: As far as I know, we can't use `ref` here because this is handled in OWL/JavaScript, where XML IDs (`ref`) are not available. That's why I used a new field to distinguish between the two dashboard instances.

Another benefit is that, in the future, if we add a new accommodation dashboard, we can simply set `x_is_accommodation_dashboard` to `true`. We won't need to make any additional code changes, since the existing logic will automatically include it.

If there's a better approach, I'm happy to change it.

booking_engine/data/spreadsheet_dashboard.xml
trcazier: Thank you for clarifying! 
Indeed, I saw that there are more instances of `spreadsheet_binary_data` than `spreadsheet_snapshot`, which is why I was asking. There are only three `spreadsheet_snapshot` (booking_engine, vineyard, 3pl) but many more `spreadsheet_binary_data`. If functionally they are not different, we should probably agree with the majority and use `spreadsheet_binary_data`, which would additionally make more sense as you explained.

booking_engine/static/src/binary/spreadsheet_dashboard/Day-to-day-Operations.osheet.json
trcazier, line 1: Wow 🤣

booking_engine/data/ir_model_fields.xml
trcazier: I see. I don't know whether a better approach exists, but I find it weird we have to do so much JS work to differentiate dashboards. I'm not saying what you did is wrong, I'm saying it's annoying that we can only do it like that...

trcazier (approved):
Impressive work, well done 💪 
LGTM 🚀

booking_engine/data/spreadsheet_dashboard.xml
vava-odoo: Thanks both for the discussion. I definitely agree we should use spreadsheet_binary_data everywhere in industry.

booking_engine/static/src/binary/spreadsheet_dashboard/Day-to-day-Operations.osheet.json
vava-odoo: You're glad now? :see_no_evil:

booking_engine/data/ir_model_fields.xml
vava-odoo: I agree with Tristan, I don't like this either. The field is not even editable by the user in any view. This looks like some magic we don't want. Could I ask you to have a round with spreadsheet team, explaining the usecase and see how they would proceed?

vava-odoo (commented):
@robodoo delegate=pgu-odoo

pgu-odoo:
@robodoo r+

booking_engine/static/src/js/dashboard.js
```diff
@@ -0,0 +1,31 @@
...
+            ]);
+            this.accommodationDashboardResId = resId;
+        })
+    }
+
+    getDashboardGroups() {
+        const groups = this.loader.getDashboardGroups();
+        groups.forEach(group => {
+            group.dashboards = group.dashboards.filter(
+                dashboard => dashboard.data.id === this.accommodationDashboardResId
```
trcazier: Since you worry about this code not being dynamic and having to extend it if we want to add another dashboard, maybe you could do something like this

```suggestion
                dashboard => this.accomodationResIds.includes(dashboard.data.id)
```

and define constants in the `setup()` that we can modify when we add another one 

```js
this.accomodationIds = ["spreadsheet_dashboard_2"]
this.accomodationResIds = null
```

then adapt `onWillStart()` to fetch the resIds

booking_engine/static/src/xml/dashboard.xml
trcazier, line 1: So in this file, you are removing everything inside these components, correct? Why is that?

trcazier (commented):
Only minor comments from me, but I had a looong discussion with SAVP about this when he learned about the way you have to implement this feature. He wants to speak with you, you two should get in touch

booking_engine/static/src/xml/dashboard.xml
amas-odoo: We are not removing the content from the original template.

When inheriting a template, the default inheritance mode is primary, which creates a copy of the inherited template and applies our modifications to that copy.

I replaced some blocks because they were not needed for our use case, so overriding them with `position="replace"` was the simplest way to remove that content from the inherited version.

trcazier: I understand that, I'm just asking why they are not needed

amas-odoo: I had a discussion regarding with @Pierre-savp , he doesn't want a search bar and sections.

vava-odoo:
task canceled

PR #2649, opened 2026-07-24
[UPG] non_profit_organization: remove computed field display_invoice_…

vava-odoo:
…template_pdf_report_id


vava-odoo:
@robodoo r+

PR #2651, opened 2026-07-24
[FIX] veterinary_clinic: remove customization for sequence generation

dhrs-odoo:
Before this commit, the sequence was generated through a server action, which contributed to the LOC count.

With this commit, the sequence generation has been moved to a computation, preventing it from being counted as custom LOC.


vava-odoo (approved):
@robodoo r+

PR #2663, opened 2026-07-24
[IMP] concert_halls: update ministock

mega-odoo:
- Remove the Inventory app.
- Reorder the app sequence.

Task-6416937


concert_halls/__manifest__.py
dhrs-odoo, line 1: As per the specs you only need to remove stock dependency `pos_stock` then why have you removed other dependencies ?

dhrs-odoo (changes requested):

concert_halls/__manifest__.py
mega-odoo: > As per the specs you only need to remove stock dependency `pos_stock` then why have you removed other dependencies ?

Removed unnecessary(extra) dependencies since `knowledge` is installed by `base_industry_data`, and `pos_event` is installed by `pos_restaurant` and `website_event`.

dhrs-odoo: Here `pos_event` seems required so auto-installed is not enough 
knowledge can be removed

dhrs-odoo (approved):
Hello @vava-odoo PR is ready for your review

vava-odoo:
@robodoo r+

PR #2665, opened 2026-07-27
[FIX] electronic_store: fix warranty field alignment

dhrs-odoo:
Before this commit, the warranty field was displayed after the stat button, making it unclear where the value should be updated.

This commit moves the warranty field below the Track Inventory field, ensuring proper alignment and improving the user experience.

Task-6387471


vava-odoo (approved):
@robodoo r+

PR #2668, opened 2026-07-27
[CLN] *: remove knowledge dependency

dhrs-odoo:
In this commit we are removing the knowledge dependency from the manifest because it is already getting installed through the base_industry_data


vava-odoo (approved):
@robodoo r+ squash

PR #2694, opened 2026-07-29
[REM] climbing_gym: remove the climbing gym industry

dhrs-odoo:
In this commit we are removing the climbing gym industry because most of the use cases are being covered in the fitness industry.

Task-6425790


vava-odoo:
@robodoo r+

PR #2696, opened 2026-07-29
[FIX] utils: handle manifest without depends key

vava-odoo:


utils.py
```diff
@@ -54,7 +54,7 @@ def get_dependencies(self, module_names: str, full_standard_list: bool = False):
                 raise Exception("manifest (%s) does not exist", manifest_file)
             with open(manifest_file, encoding="utf-8") as f:
                 manifest = ast.literal_eval(f.read())
-            for dep in manifest.get('depends'):
+            for dep in manifest.get('depends', []):
```
trcazier: Nitpicking: shouldn't we add base by default then?

trcazier (approved):
LGTM 🚀

vava-odoo:
@robodoo delegate=trcazier

trcazier:
@robodoo r+

PR #2699, opened 2026-07-30
[FIX] construction_developer: align unit price and margin formulas

chga-odoo:
- Redefine unit margin as a percentage of the selling price
- Update the inverse unit price calculation to keep both formulas mathematically consistent
- Update affected tests to validate the new behavior

Task: 6273532


vava-odoo (approved):
@robodoo r+

PR #2705, opened 2026-07-30
[IMP] booking_engine: prevent bookings years in advance

aycha-odoo:
Rental businesses generally do not want to accept bookings several years
ahead as availability and pricing may change over time.

Add a default pricelist rule that sets the booking price to 0 beyond one
year in the future. Combined with the "Hide Add To Cart" website setting,
this prevents customers from booking dates outside the supported booking
window.

Update industry demo websites and knowledge articles to document and
support this configuration.

task-6246347

### DIFF
--- a/hotel/demo/website/website.xml
+++ b/hotel/demo/website/website.xml
@@ -10,6 +10,7 @@
         <field name="show_line_subtotals_tax_selection">tax_included</field>
         <field name="logo" type="base64" file="hotel/static/src/img/content/website/logo.webp"/>
         <field name="favicon" type="base64" file="hotel/static/src/img/content/website/favicon.png"/>
+        <field name="prevent_sale" eval="True"/>
     </record>
 
     <function name="write" model="ir.model.data">
@@ -19,5 +20,9 @@
             ('model', '=', 'website'),
         ]"/>
         <value eval="{'res_id': ref('website_industry')}"/>
+    </function>
+    <function name="write" model="product.pricelist">
+        <value model="product.pricelist" search="[('id', '=', ref('booking_engine.product_pricelist_1'))]"/>
+        <value eval="{'website_id': ref('website_industry')}"/>
     </function>
 </odoo>
--- a/booking_engine/data/product_pricelist_item.xml
+++ b/booking_engine/data/product_pricelist_item.xml
@@ -36,4 +36,15 @@
     <field name="percent_price">-50.0</field>
     <field name="x_name">Christmas</field>
   </record>
+  <record id="pricelist_item_block_future_booking" model="product.pricelist.item">
+    <field name="pricelist_id" ref="product_pricelist_1" />
+    <field name="applied_on">2_product_category</field>
+    <field name="display_applied_on">2_product_category</field>
+    <field name="categ_id" ref="product_category_6" />
+    <field name="compute_price">fixed</field>
+    <field name="fixed_price">0.0</field>
+    <field name="date_start" eval="datetime.now() + relativedelta(years=1)" />
+    <field name="date_end" eval="datetime.now() + relativedelta(years=30)" />
+    <field name="x_name">Booking Blocker</field>
+  </record>
 </odoo>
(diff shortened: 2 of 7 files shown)


campsite/demo/website.xml
```diff
@@ -13,5 +13,6 @@
     <field name="shop_ppr">3</field>
     <field name="shop_gap">10px</field>
     <field name="product_page_image_width">66_pc</field>
+    <field name="prevent_sale" eval="True"/>
```
vava-odoo: I'd move this to res_config_setting in booking_engin/data/, so that
- it is active in production db
- no need to duplicate in all industry

hotel/demo/website/website.xml
```diff
@@ -20,4 +21,8 @@
         ]"/>
         <value eval="{'res_id': ref('website_industry')}"/>
     </function>
+    <function name="write" model="product.pricelist">
+        <value model="product.pricelist" search="[('id', '=', ref('booking_engine.product_pricelist_1'))]"/>
+        <value eval="{'website_id': ref('website_industry')}"/>
+    </function>
```
vava-odoo, line 27: This will probably be necessary, since we apply the pricelist to a new website

vava-odoo:
@robodoo delegate=frva-odoo
@frva-odoo

campsite/demo/website.xml
aycha-odoo: Thanks for the suggestion!

I tried moving this to `booking_engine/data/` to avoid the duplication, but ran into a technical limitation.

`prevent_sale` on `res.config.settings` is defined as:

```python
prevent_sale = fields.Boolean(
    related='website_id.prevent_sale',
    readonly=False,
)
```

This is a website-specific related field, meaning it reads/writes on whichever `website_id` is active in the settings context when `execute()` is called.

Because of that, setting it through a `<record>` + `execute()` in `booking_engine/data/res_config_settings.xml` doesn't have any effect during data loading, since no specific website is available in the context at that point.

I also tried using a:

```xml
<function name="write" model="website">
```

to write directly on website records , but in my testing that wasn't applied either.

Since each industry module creates its own website record, setting `prevent_sale="True"` directly on that website in `demo/website.xml` is the only approach I've found that works.

That said, I may be missing a cleaner centralized solution, so if you have another approach in mind, I'm happy to explore it. :)

frva-odoo: I think what Vava meant here was to apply it like you did before, but in the booking_engine module so we don't repeat it in the 4 dependent modules.
As it is applied to the default website that one is already available before, so we can write that value on it:
```xml
<function name="write" model="website">
    <value eval="[ref('website.default_website')]"/>
    <value eval="{'prevent_sale': True}"/>
</function>
``` 
should work and is better if the website exists before executing the file (should be the case), otherwise:
```xml
<record id="website.default_website" model="website" forcecreate="1">
    <field name="name">My Website</field>
    <field name="configurator_done" eval="True"/>
    <field name="prevent_sale" eval="True"/>
</record>
``` 
This takes the existing record default_website if it finds it or creates it if it does not exist, and assign the values to it.
Put this in the `<data noupdate="1">` in booking_engine/data/res_config_settings.xml, this way it will not change the field value each time a user updates the module (only when it is first installed)

hotel/demo/website/website.xml
```diff
@@ -10,6 +10,7 @@
         <field name="show_line_subtotals_tax_selection">tax_included</field>
         <field name="logo" type="base64" file="hotel/static/src/img/content/website/logo.webp"/>
         <field name="favicon" type="base64" file="hotel/static/src/img/content/website/favicon.png"/>
+        <field name="prevent_sale" eval="True"/>
```
frva-odoo, line 13: This however should stay here, as it is not the same website record. Here we don't modify default_website but create a new one website_industry instead, so keep prevent_sale true here

frva-odoo (changes requested):
Hello @aycha-odoo, taking over the review as Vava is off 🏖️ 
Great work for this PR, only one comment on the same part as Vava. Don't hesitate if you have any question 😄

frva-odoo (approved):
Perfect, good job @aycha-odoo 👏

frva-odoo:
@robodoo r+

PR #2733, opened 2026-08-04
[IMP] textile_manufacturing: add a setting for product conversion module

mega-odoo:
- Add fields to install or uninstall the product conversion module in inventory settings (disabled by default).

- Update the related Knowledge article.

Task ID - 6357143

### DIFF
--- a/textile_manufacturing/__manifest__.py
+++ b/textile_manufacturing/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Textile Manufacturing',
-    'version': '1.5',
+    'version': '1.6',
     'category': 'Supply Chain',
     'author': 'Odoo S.A.',
     'depends': [
@@ -9,15 +9,16 @@
         'documents_project',
         'hr_timesheet',
         'mrp_subcontracting',
-        'product_conversion',
         'purchase_mrp',
         'quality_mrp_workorder',
         'sale_project',
+        'web_studio',
         'website_crm',
         'website_sale_stock',
     ],
     'data': [
         'data/res_config_settings.xml',
+        'data/res_config_settings_views.xml',
         'data/ir_ui_menu.xml',
         'data/ir_attachment_pre.xml',
         'data/documents_document.xml',
@@ -75,6 +76,7 @@
     ],
     'license': 'OEEL-1',
     'cloc_exclude': [
+        'data/res_config_settings.xml',
         'data/knowledge_article.xml',
         'demo/website_view.xml',
     ],
--- a/textile_manufacturing/__manifest__.py
+++ b/textile_manufacturing/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Textile Manufacturing',
-    'version': '1.8',
+    'version': '1.9',
     'category': 'Supply Chain',
     'author': 'Odoo S.A.',
     'depends': [
@@ -9,16 +9,17 @@
         'documents_project',
         'hr_timesheet',
         'mrp_subcontracting',
-        'product_conversion',
         'purchase_mrp',
         'quality_mrp_workorder',
         'sale_product_matrix',
         'sale_project',
+        'web_studio',
         'website_crm',
         'website_sale_stock',
     ],
     'data': [
         'data/res_config_settings.xml',
+        'data/res_config_settings_views.xml',
         'data/ir_ui_menu.xml',
         'data/ir_attachment_pre.xml',
         'data/documents_document.xml',
@@ -76,6 +77,7 @@
     ],
     'license': 'OEEL-1',
     'cloc_exclude': [
+        'data/res_config_settings.xml',
         'data/knowledge_article.xml',
         'demo/website_view.xml',
     ],
--- a/textile_manufacturing/data/product_attribute.xml
+++ b/textile_manufacturing/data/product_attribute.xml
@@ -2,6 +2,5 @@
 <odoo noupdate="1">
     <record id="product_attribute_1" model="product.attribute">
         <field name="name">Quality Level</field>
-        <field name="x_is_convert_attribute">True</field>
     </record>
 </odoo>
--- a/textile_manufacturing/data/res_config_settings.xml
+++ b/textile_manufacturing/data/res_config_settings.xml
@@ -1,10 +1,102 @@
 <?xml version='1.0' encoding='UTF-8'?>
-<odoo noupdate="1">
-    <record id="res_config_setting_textile_manufacturing" model="res.config.settings">
-        <field name="group_mrp_routings" eval="1"/>
-        <field name="group_project_stages" eval="1"/>
+<odoo>
+    <data noupdate="1">
+        <record id="res_config_setting_textile_manufacturing" model="res.config.settings">
+            <field name="group_mrp_routings" eval="1"/>
+            <field name="group_project_stages" eval="1"/>
+            <field name="group_stock_production_lot" eval="1"/>
+        </record>
+        <function name="execute" model="res.config.settings">
+            <value eval="[ref('res_config_setting_textile_manufacturing')]"/>
+        </function>
+    </data>
+
+    <record id="x_res_company_field_module_product_conversion" model="ir.model.fields">
+        <field name="ttype">boolean</field>
+        <field name="field_description">Product Conversion</field>
+        <field name="name">x_module_product_conversion</field>
+        <field name="model_id" ref="base.model_res_company"/>
     </record>
-    <function name="execute" model="res.config.settings">
-        <value eval="[ref('res_config_setting_textile_manufacturing')]"/>
-    </function>
+    <record id="x_setting_field_module_product_conversion" model="ir.model.fields">
+        <field name="ttype">boolean</field>
+        <field name="related">company_id.x_module_product_conversion</field>
+        <field name="name">x_module_product_conversion</field>
+        <field name="model_id" ref="base_setup.model_res_config_settings"/>
+    </record>
+    <!-- Action Install Product Conversion Module From Settings -->
+    <record id="action_activate_product_conversion" model="ir.actions.server">
+        <field name="code"><![CDATA[
+if not env.context.get('prevent_product_conversion_install_recursion'):
+    module_ref = env['ir.module.module']
+    # First, import the Product Conversion industry module because it is not available in ir_module_module.
+    import_action = module_ref.with_context(module_name='product_conversion').button_immediate_install_app()
+    env['base.import.module'].browse(import_action['res_id']).with_context(prevent_product_conversion_install_recursion=True).import_module()
+    module = module_ref._get('product_conversion')
+    if module and module.state == 'uninstalled': module.button_immediate_install()
+]]>        </field>
+        <field name="model_id" ref="base_setup.model_res_config_settings"/>
+        <field name="state">code</field>
+        <field name="name">Install Product Conversion</field>
+    </record>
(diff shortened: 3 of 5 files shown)


textile_manufacturing/data/res_config_settings.xml
```diff
@@ -1,10 +1,102 @@
...
+
+    <record id="x_res_company_field_module_product_conversion" model="ir.model.fields">
+        <field name="ttype">boolean</field>
+        <field name="field_description">Product Conversion</field>
+        <field name="name">x_module_product_conversion</field>
+        <field name="model_id" ref="base.model_res_company"/>
+    </record>
+    <record id="x_setting_field_module_product_conversion" model="ir.model.fields">
+        <field name="ttype">boolean</field>
+        <field name="field_description">Product Conversion</field>
```
pgu-odoo, line 16:
```suggestion
```

textile_manufacturing/data/res_config_settings.xml
```diff
@@ -1,10 +1,102 @@
...
+    </record>
+    <!-- Action Install Product Conversion Module From Settings -->
+    <record id="action_activate_product_conversion" model="ir.actions.server">
+        <field name="code"><![CDATA[
+if not env.context.get('prevent_product_conversion_install_recursion'):
+    module_ref = env['ir.module.module']
+    import_action = module_ref.with_context(module_name='product_conversion').button_immediate_install_app()
+    env['base.import.module'].browse(import_action['res_id']).with_context(prevent_product_conversion_install_recursion=True).import_module()
+    module = module_ref._get('product_conversion')
+    if module and module.state == 'uninstalled': module.button_immediate_install()
```
pgu-odoo: Why there is a need of this check? when we already called import_module()?
could also be restructured like - 


```
if not env.context.get('prevent_product_conversion_install_recursion'):
    module = env['ir.module.module']._get('product_conversion')

    if module and module.state == 'uninstalled':
        env['base.import.module'].browse(env['ir.module.module']
            .with_context(module_name='product_conversion')
            .button_immediate_install_app()).with_context(
            prevent_product_conversion_install_recursion=True
        ).import_module()
```

textile_manufacturing/data/res_config_settings_views.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="name">res.config.settings.view.form.inherit.textile.manufacturing</field>
+        <field name="model">res.config.settings</field>
+        <field name="active" eval="True"/>
+        <field name="inherit_id" ref="stock.res_config_settings_view_form"/>
+        <field name="arch" type="xml">
+            <block name="operations_setting_container" position="inside">
+                <setting id="product_conversion" string="Product Conversion">
+                    <field name="x_module_product_conversion"/>
+                    <span class="text-muted">
+                        Easily convert products and manage quality levels. (See <a href="https://apps.odoo.com/apps/modules/saas-19.3/product_conversion" target="_blank">documentation</a>). Activating this feature runs additional code and may affect your subscription billing.
```
pgu-odoo: static link for saas-19.3?

pgu-odoo (commented):
@mega-odoo a few remarks are here. Good job although.

pgu-odoo:

textile_manufacturing/data/res_config_settings.xml
```diff
@@ -1,11 +1,102 @@
...
+        <function name="execute" model="res.config.settings">
+            <value eval="[ref('res_config_setting_textile_manufacturing')]"/>
+        </function>
+    </data>
+
+    <record id="x_res_company_field_module_product_conversion" model="ir.model.fields">
+        <field name="ttype">boolean</field>
+        <field name="field_description">Product Conversion</field>
+        <field name="name">x_module_product_conversion</field>
+        <field name="model_id" ref="base.model_res_company"/>
```
vava-odoo: I'd rather work with an ir.config.parameter, because I don't know how this would deal with multi-company setup

textile_manufacturing/data/res_config_settings.xml
```diff
@@ -1,11 +1,102 @@
...
+    <record id="x_setting_field_module_product_conversion" model="ir.model.fields">
+        <field name="ttype">boolean</field>
+        <field name="related">company_id.x_module_product_conversion</field>
+        <field name="name">x_module_product_conversion</field>
+        <field name="model_id" ref="base_setup.model_res_config_settings"/>
+    </record>
+    <!-- Action Install Product Conversion Module From Settings -->
+    <record id="action_activate_product_conversion" model="ir.actions.server">
+        <field name="code"><![CDATA[
+if not env.context.get('prevent_product_conversion_install_recursion'):
```
vava-odoo: the ICP should remove this potential recursion issue

textile_manufacturing/data/res_config_settings.xml
```diff
@@ -1,11 +1,102 @@
...
+    module_ref = env['ir.module.module']
+    # First, import the Product Conversion industry module because it is not available in ir_module_module.
+    import_action = module_ref.with_context(module_name='product_conversion').button_immediate_install_app()
+    env['base.import.module'].browse(import_action['res_id']).with_context(prevent_product_conversion_install_recursion=True).import_module()
+    module = module_ref._get('product_conversion')
+    if module and module.state == 'uninstalled': module.button_immediate_install()
```
vava-odoo: this is impossible: uninstalling an imported module remove the ir.module.module entry

textile_manufacturing/data/res_config_settings.xml
```diff
@@ -1,11 +1,102 @@
...
+]]>        </field>
+        <field name="model_id" ref="base_setup.model_res_config_settings"/>
+        <field name="state">code</field>
+        <field name="name">Install Product Conversion</field>
+    </record>
+    <!--Action Uninstall Product Conversion Module From Settings -->
+    <record id="action_deactivate_product_conversion" model="ir.actions.server">
+        <field name="code"><![CDATA[
+module = env['ir.module.module']._get('product_conversion')
+if module and module.state == 'installed': module.button_immediate_uninstall()
```
vava-odoo:
```suggestion
if module: module.button_immediate_uninstall()
```

textile_manufacturing/data/product_attribute.xml
```diff
@@ -2,6 +2,5 @@
 <odoo noupdate="1">
     <record id="product_attribute_1" model="product.attribute">
         <field name="name">Quality Level</field>
-        <field name="x_is_convert_attribute">True</field>
```
vava-odoo, line 5: @lpieodoo this is the only attribute that showcases this conversion feature. Are all (demo) data and flows still relevant without this? Couldn't we imagine to add this when activating the product_conversion (maybe only if demo is active)?

lpieodoo: Yes, even without this attribute the flow remains relevant. So we can add this when we activate the setting

textile_manufacturing/__manifest__.py
```diff
@@ -9,15 +9,16 @@
         'documents_project',
         'hr_timesheet',
         'mrp_subcontracting',
-        'product_conversion',
         'purchase_mrp',
         'quality_mrp_workorder',
         'sale_project',
+        'web_studio',
```
vava-odoo, line 16: as discussed on the task
```suggestion
```

vava-odoo (commented):
@frva-odoo for you :-)

textile_manufacturing/data/res_config_settings.xml
```diff
@@ -1,11 +1,79 @@
...
+        <field name="filter_domain">[('name', '=', 'product_conversion')]</field>
+        <field name="trigger">on_create_or_write</field>
+    </record>
+    <record id="action_enable_product_conversion_attribute" model="ir.actions.server">
+        <field name="name">Enable Product Conversion Attribute</field>
+        <field name="model_id" ref="base.model_ir_model_fields"/>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+attribute = env.ref('textile_manufacturing.product_attribute_1', raise_if_not_found=False)
+if attribute: attribute.x_is_convert_attribute = True
```
frva-odoo:
```suggestion
if attribute: attribute['x_is_convert_attribute'] = True
```

textile_manufacturing/data/res_config_settings.xml
```diff
@@ -1,11 +1,79 @@
...
+        <field name="name">x_module_product_conversion</field>
+        <field name="store" eval="True"/>
+        <field name="model_id" ref="base_setup.model_res_config_settings"/>
+        <field name="compute"><![CDATA[
+self['x_module_product_conversion'] = self.env['ir.config_parameter'].get_bool('textile_manufacturing.x_setting_product_conversion', False)
+]]>        </field>
+    </record>
+    <record id="action_product_conversion" model="ir.actions.server">
+        <field name="code"><![CDATA[
+if not (env.context.get('install_module') == 'product_conversion' and env.context.get('install_mode')):
```
frva-odoo, line 26: Why do we need this check about install_module and install_mode ?

frva-odoo (commented):
The setting is well done, great work @mega-odoo 👏 2 small remarks from me

textile_manufacturing/data/res_config_settings.xml
mega-odoo: > Why do we need this check about install_module and install_mode ?


### When the user enables **Product Conversion** from Settings

We face a recursion issue when the user enables **Product Conversion** from Settings. But this could be avoided by using `ir.config_parameter`. 

1. Enabling **Product Conversion** from settings starts the import and installation of the `product_conversion` module.
2. During the module installation, `res.config.settings` records are loaded/created again.
3. Since `res.config.settings` is a transient model, `x_module_product_conversion` is evaluated as `False`.Because it cannot be stored due to the transient model.
4. This triggers the **Activate Product Conversion** base automation, which attempts to uninstall the `product_conversion` module while it is still in the installation process.

### When the user installs **Product Conversion** directly from Apps

In this case, relying only on `ir.config_parameter` is not sufficient because the installation is triggered directly from Apps.

1. The user starts installing the `product_conversion` module from Apps.
2. During the installation, `res.config.settings` records are loaded/created again.
3. Since `res.config.settings` is transient, `x_module_product_conversion`  is evaluated as `False`.
4. This triggers the **Activate Product Conversion** base automation, which attempts to uninstall the module while its installation is still in progress.

To handle both scenarios, we use a value in `env.context` to identify when the **Product Conversion** module is currently being installed. The automation checks this context value and skips the uninstall while the installation process is still running, and preventing the recursion issue.


In short, 

Handling this recursion issue using only `ir.config_parameter` may be difficult, especially when the module is installed directly from Apps.

We need to use an `env.context` value to detect when the **Product Conversion** module is currently being installed and to prevent this recursion issue.

Alternatively, we could store the value of `x_module_product_conversion` on `res.company` to avoid this recursion issue.

frva-odoo: Okay I see, this seems to work well so we can keep this in this module. Handling the uninstallation issue by a context seems pretty good 👍

mega-odoo:
> The setting is well done, great work @mega-odoo 👏 2 small remarks from me

Changes have been made according to your suggestion. Thank you :)

textile_manufacturing/__manifest__.py
vava-odoo: I reverted this since base_automation brings the same issue as web_studio.
And I also edited the link in the setting.

vava-odoo (approved):
@robodoo r+

PR #2745, opened 2026-08-06
[FIX] booking_engine, hotel: restore functional housekeeping demo data

aycha-odoo:
The housekeeping demo data was no longer usable out of the box, making it difficult to validate the housekeeping planning flow in demo environments.

Several assumptions in the demo setup depended on the database creation time and on employee calendars being configured in a specific way. As a result, demo shifts could end up with unexpected durations, employees were not properly available for automatic planning, and some housekeeping planning indicators were not computed correctly.

This commit makes the demo data deterministic and aligned with the expected housekeeping workflow by:

- normalizing demo slot datetimes to fixed hour/minute boundaries to avoid database setup time affecting shift durations;
- configuring demo housekeeping employees with planning-friendly availability so shifts can be assigned and scheduled correctly;
- fixing the allocation computation used by housekeeping resources to ensure values are initialized and aggregated properly;
- aligning the Booking planning menu/action label with the standard Planning terminology used throughout the application.

The goal is to provide a fully functional demo environment where housekeeping planning, assignment and workload visualization behave consistently across databases.

task-6425559

### DIFF
--- a/campsite/demo/planning_slot_post.xml
+++ b/campsite/demo/planning_slot_post.xml
@@ -1,4 +1,250 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <odoo>
+    <data noupdate="1">
+        <record id="planning_slot_26" model="planning.slot">
+            <field name="start_datetime" eval="DateTime.now().replace(hour=7, minute=0, second=0, microsecond=0) + relativedelta(days=11)"/>
+            <field name="end_datetime" eval="DateTime.now().replace(hour=11, minute=0, second=0, microsecond=0) + relativedelta(days=11)"/>
+            <field name="template_id" ref="booking_engine.planning_slot_template_1"/>
+            <field name="role_id" ref="booking_engine.house_keeping_planning_role"/>
+            <field name="x_nights">1</field>
+        </record>
+        <record id="planning_slot_25" model="planning.slot">
+            <field name="start_datetime" eval="DateTime.now().replace(hour=7, minute=0, second=0, microsecond=0) + relativedelta(days=11)"/>
+            <field name="end_datetime" eval="DateTime.now().replace(hour=9, minute=0, second=0, microsecond=0) + relativedelta(days=11)"/>
+            <field name="template_id" ref="booking_engine.planning_slot_template_2"/>
+            <field name="role_id" ref="booking_engine.house_keeping_planning_role"/>
+            <field name="x_nights">1</field>
+        </record>
+        <record id="planning_slot_24" model="planning.slot">
+            <field name="start_datetime" eval="DateTime.now().replace(hour=7, minute=0, second=0, microsecond=0) + relativedelta(days=11)"/>
+            <field name="end_datetime" eval="DateTime.now().replace(hour=11, minute=0, second=0, microsecond=0) + relativedelta(days=11)"/>
+            <field name="template_id" ref="booking_engine.planning_slot_template_1"/>
+            <field name="role_id" ref="booking_engine.house_keeping_planning_role"/>
+            <field name="x_nights">1</field>
+        </record>
+        <record id="planning_slot_29" model="planning.slot">
+            <field name="start_datetime" eval="DateTime.now().replace(hour=7, minute=0, second=0, microsecond=0) + relativedelta(days=10)"/>
+            <field name="end_datetime" eval="DateTime.now().replace(hour=11, minute=0, second=0, microsecond=0) + relativedelta(days=10)"/>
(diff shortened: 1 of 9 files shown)


campsite/demo/planning_slot_post.xml
```diff
@@ -1,4 +1,257 @@
...
+            <field name="role_id" ref="booking_engine.house_keeping_planning_role"/>
+            <field name="x_nights">1</field>
+        </record>
+        <record id="planning_slot2" model="planning.slot">
+            <field name="start_datetime" eval="DateTime.now().replace(hour=7, minute=0, second=0, microsecond=0) - relativedelta(days=1)"/>
+            <field name="end_datetime" eval="DateTime.now().replace(hour=9, minute=0, second=0, microsecond=0) - relativedelta(days=1)"/>
+            <field name="template_id" ref="booking_engine.planning_slot_template_2"/>
+            <field name="role_id" ref="booking_engine.house_keeping_planning_role"/>
+            <field name="x_nights">1</field>
+        </record>
```
pgu-odoo, line 213: I think exact same data? 🤔

pgu-odoo (commented):
@aycha-odoo good work! Just a remark and don't forget to version bump.

abdrahmanrashed (approved):
Great Work!

hotel/demo/planning_slot_post.xml
```diff
@@ -247,8 +240,8 @@
         <function name="action_send" model="planning.send">
             <value model="planning.send" eval="[obj().env['planning.send'].create({
                 'employee_ids': [(6, 0, [ref('booking_engine.hr_employee_2'), ref('booking_engine.hr_employee_3'), ref('booking_engine.hr_employee_4')])],
-                'start_datetime': DateTime.now() - relativedelta(days=3),
-                'end_datetime': DateTime.now() + relativedelta(days=4),
+                'start_datetime': DateTime.now().replace(hour=0, minute=0, second=0, microsecond=0) - relativedelta(days=3),
+                'end_datetime': DateTime.now().replace(hour=23, minute=59, second=59, microsecond=0) + relativedelta(days=4),
```
abdrahmanrashed: really small thing, it would probably be better to set microseconds to max value as well
```suggestion
                'end_datetime': DateTime.now().replace(hour=23, minute=59, second=59, microsecond=999999) + relativedelta(days=4),
```

booking_engine/data/ir_ui_menu.xml
```diff
@@ -29,4 +29,10 @@
       </menuitem>
     </menuitem>
   </menuitem>
+  <record id="planning.planning_menu_planning" model="ir.ui.menu" forcecreate="0">
+    <field name="active" eval="False"/>
+  </record>
+  <record id="planning.planning_menu_my_planning" model="ir.ui.menu" forcecreate="0">
+    <field name="active" eval="False"/>
+  </record>
```
vava-odoo: why this?

hotel/demo/planning_slot_post.xml
vava-odoo: why? And why not in campsite?

aycha-odoo: I think setting it to 999999 instead of 0 ensures that we capture the absolute, exact end of the day. If we leave it at 0, there is technically still 0.999999 seconds left in the day where a record could theoretically fall outside the window.

Also, you are completely right about campsite! I have just updated campsite/demo/planning_slot_post.xml to include this fix as well so the demo data is consistent across both modules.

booking_engine/data/ir_ui_menu.xml
aycha-odoo: @Pierre-savp didn't want this extra planning in the planning app
image

vava-odoo: @Pierre-savp sure about this?

Pierre-savp: @vava-odoo 
was mentionning, this is not in standard, so why would we have this as a landing? 
we don't want planning removing to be custom indeed, we just want planning to be behaving like standard RB landing on schedule 

standard RB:
image

vava-odoo: To get back to standard, we should not deactivate more standard things...
What I mean is that we should modify the existing customization instead. I bet there are some views that should be primary instead of extension.

vava-odoo (approved):
just a conflict to solve

vava-odoo:
@robodoo override=ci/industry
@robodoo r+

PR #2750, opened 2026-08-10
[IMP] museum: adapt frontdesk and ticketing workflow

chga-odoo:
- Add the Frontdesk app.
- Remove the Inventory app.
- Resequence apps on the dashboard.
- Remove the 'Supporter Membership' subscription product.
- Rename and remove obsolete CRM levels.
- Create a new frontdesk station named 'Members Entrance'.
- Add a ticket to the event configuration.
- Remove the custom pricelist.
- Update the knowledge article to reflect the new workflow.

Task: 6431446


museum/__manifest__.py
```diff
@@ -1,21 +1,23 @@
...
     'name': 'Museum',
-    'version': '1.2',
+    'version': '2.0',
     'category': 'Services',
     'author': 'Odoo S.A.',
     'depends': [
         'base_industry_data',
         'calendar',
         'documents_sign',
+        'frontdesk',
```
dhrs-odoo:
```suggestion
        'frontdesk_partnership',
```

museum/__manifest__.py
```diff
@@ -1,21 +1,23 @@
 {
     'name': 'Museum',
-    'version': '1.2',
+    'version': '2.0',
```
dhrs-odoo: @vava-odoo do we need major bump over here ?
As it's not that major update

museum/data/product_pricelist.xml
dhrs-odoo, line 1: Remove this default pricelist as it will be created automatically

dhrs-odoo (changes requested):
Hello @chga-odoo 
Great work 👏 
Just two minor comments 🤏 
Thanks

museum/__manifest__.py
chga-odoo: can you please explain?

dhrs-odoo: Since we previously only had the `partnership` module and no `frontdesk` module, adding `partnership` as an individual dependency made sense.

However, with this task, we are now adding the `frontdesk` dependency as well. In that case, there is no need to keep both `frontdesk` and `partnership` as separate dependencies.We could use `frontdesk_partnership` instead, since it will eventually install both modules

dhrs-odoo (approved):
@vava-odoo PR is Ready for your review

museum/data/product_pricelist_item.xml
vava-odoo: seems like this one is no more used, right?

museum/demo/stock_quant.xml
```diff
@@ -1,92 +0,0 @@
-<?xml version='1.0' encoding='UTF-8'?>
-<odoo noupdate="1">
-  <record id="stock_quant_1" model="stock.quant">
-    <field name="product_id" ref="product_product_13"/>
-    <field name="location_id" ref="stock.stock_location_stock"/>
-    <field name="inventory_quantity_auto_apply">1.0</field>
```
vava-odoo, line 6: wouldn't it still make sense to add the quantity in the mini-stock way?

museum/__manifest__.py
vava-odoo: minor is enough, yes
```suggestion
    'version': '1.3',
```

vava-odoo (approved):

vava-odoo:
@robodoo r+

PR #2761, opened 2026-08-17
[FIX] condominium: fix invoicing and distribution key issues

dhrs-odoo:
In this commit, we have fixed several issues
- Updated the invoicing address on split SOs to use the contact/owner instead of the previous contact
- Restricted distribution key domains from SOs and event calendars to share keys
- Renamed "View Report" to "Preview Report" in calendar events.
- Fixed the distribution key demo data

Task-6469472

### DIFF
--- a/condominium/__manifest__.py
+++ b/condominium/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Property Owner Association',
-    'version': '3.4',
+    'version': '3.5',
     'category': 'Services',
     'depends': [
         'account_check_printing',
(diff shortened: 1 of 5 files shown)


condominium/data/ir_model_fields.xml
```diff
@@ -577,6 +577,13 @@ for record in self: record['x_companies'] = [(6, 0, record.x_property_ids.x_cond
     <field name="relation">res.company</field>
     <field name="tracking">1</field>
   </record>
+  <record id="field_distribution_key_building" model="ir.model.fields">
```
vava-odoo: The goal is not to change the functional aspect here. The goal is to improve demo data. 
So what used to work in demo by calling the server action doesn't anymore. You'll need to reference all properties for the specific wing.
We're not adding any field in this task

vava-odoo (changes requested):

vava-odoo (approved):
@robodoo r+

vava-odoo:
@robodoo override=ci/industry

PR #2763, opened 2026-08-17
[IMP] industry_real_estate: add meter invoicing

abdrahmanrashed:
tasks-6318046

### DIFF
--- a/condominium/data/ir_actions_server.xml
+++ b/condominium/data/ir_actions_server.xml
@@ -18,65 +18,6 @@ for owner in env['res.partner'].search([('x_companies', '=', record.company_id.i
 record.write({'state': 'cancel'})
 ]]>        </field>
     </record>
-    <record id="ir_actions_server_distribute_costs" model="ir.actions.server">
-        <field name="name">Condominium: Distribute Costs</field>
-        <field name="binding_model_id" ref="account.model_account_move"/>
-        <field name="model_id" ref="account.model_account_move"/>
-        <field name="state">code</field>
-        <field name="code"><![CDATA[
-def findClosestMeterReading(date, meter_readings):
-    closest_mr = meter_readings[0]  # Assume first is a good start
-    if date:
-        for mr in meter_readings:
-            if abs((date - mr.x_date).days) < abs((date - closest_mr.x_date).days): closest_mr = mr
-    else:  # Case when owner is still current owner, end date is False, then take the latest readings available
-        for mr in meter_readings:
-            if closest_mr.x_date < mr.x_date: closest_mr = mr
-    return closest_mr
-
-if not record.x_distribution_key:
-    record.invoice_line_ids.write({'analytic_distribution': {}})
-else:
-    analytic_distribution = {}
-    periods_valid = bool(record.x_period_start) == bool(record.x_period_end)
-    if record.x_distribution_key.x_based_on == "Shares" and periods_valid:
-        for ratio in record.x_distribution_key.x_ratio_ids.filtered(lambda l: l.x_ratio != 0):
-            for account in ratio.x_property_ratios.x_account_ids:
-                if record.x_period_start and record.x_period_end:
-                    if account.x_start_date <= record.x_period_end and (not account.x_end_date or record.x_period_start <= account.x_end_date):
-                        # compute overlap start once
-                        overlap_start = max(account.x_start_date, record.x_period_start)
-                        # compute overlap end once
-                        overlap_end = account.x_end_date if account.x_end_date else record.x_period_end
-                        analytic_distribution[account.id] = ((overlap_end - overlap_start).days + 1) * ratio.x_ratio
-                elif account.partner_id.id == ratio.x_owner.id: analytic_distribution[account.id] = ratio.x_ratio
-    # Meter distribution
-    if record.x_distribution_key.x_based_on == "Meter Readings" and record.x_period_start and record.x_period_end:
-        for property in env['x_property'].search([('x_company_id', 'in', record.x_distribution_key.x_company_id.id)]):
-            meter_reading_ids = property.x_meter_reading_ids.filtered(lambda mr: mr.x_meter_id.id == record.x_distribution_key.x_meter.id)
-            if len(meter_reading_ids) >= 2:
-                billstart = findClosestMeterReading(record.x_period_start, meter_reading_ids)
-                billend = findClosestMeterReading(record.x_period_end, meter_reading_ids)
-                for account in property.x_account_ids:
-                    if not account.x_start_date or account.x_start_date > record.x_period_end or (account.x_end_date and account.x_end_date <= record.x_period_start): continue
-                    mrostart = findClosestMeterReading(account.x_start_date, meter_reading_ids)
-                    mrostart_date = mrostart.x_date
-                    billend_date = billend.x_date
-                    if mrostart_date < billstart.x_date: mrostart = billstart
-                    if account.x_end_date:
-                        mroend = findClosestMeterReading(account.x_end_date, meter_reading_ids)
-                        if billend_date < mroend.x_date: mroend = billend
-                    else: mroend = billend
-                    if not (billend_date < mrostart_date or billend_date < mroend.x_date) and mrostart_date != mroend.x_date:
-                        analytic_distribution[account.id] = abs(mroend.x_quantity - mrostart.x_quantity)
-    # Outcome
-    if (total_ratio := sum(analytic_distribution.values())):
-        for key in analytic_distribution: analytic_distribution[key] *=  100 / total_ratio
-        for aml in record.invoice_line_ids: aml['analytic_distribution'] = analytic_distribution
-    else:
-        for aml in record.invoice_line_ids: aml['analytic_distribution'] = {}
-]]></field>
-    </record>
     <record id="ir_act_server_confirm_action" model="ir.actions.server">
         <field name="name">Confirm</field>
         <field name="state">code</field>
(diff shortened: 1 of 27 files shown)


property_assets_distribution/data/base_automation.xml
```diff
@@ -21,4 +21,18 @@
         <field name="trigger">on_create_or_write</field>
         <field name="trigger_field_ids" eval="[(6, 0, [ref('field_stake_holder_partner_id'), ref('field_stake_holder_property_id')])]"/>
     </record>
+    <record id="automation_call_distribute_costs_on_update" model="base.automation">
+        <field name="name">On Distribution input update</field>
+        <field name="model_id" ref="account.model_account_move"/>
+        <field name="action_server_ids" eval="[(6, 0, [ref('property_assets_distribution.ir_actions_server_distribute_costs')])]"/>
```
vava-odoo: no need for self module ref
```suggestion
        <field name="action_server_ids" eval="[(6, 0, [ref('ir_actions_server_distribute_costs')])]"/>
```

condominium/data/ir_rule.xml
```diff
@@ -12,12 +12,12 @@
     </record>
     <record id="ratios_multi_company" model="ir.rule">
         <field name="name">Ratios - Multi-Company</field>
-        <field name="model_id" ref="ratios_model"/>
+        <field name="model_id" ref="property_assets_distribution.ratios_model"/>
```
vava-odoo: probably to move to common module as well, no?

property_assets_distribution/data/ir_ui_view.xml
```diff
@@ -293,4 +304,137 @@
...
+            <group name="distribution_group_left" string="Distribution">
+              <field name="x_distribution_key"
+                     options="{'no_create': true}"
+              />
+              <field name="x_period_start"
+                     required="x_distribution_key or x_based_on == 'Meter Readings'"
+                     help="Used to distribute costs by meter; closest meter reading date used"
+              />
+              <field name="x_period_end"
+                     help="Used to distribute costs by meter; closest meter reading date used"
```
vava-odoo, line 331: missing required

property_assets_distribution/data/ir_model_fields.xml
```diff
@@ -431,4 +446,140 @@ for record in self: record['x_account_ids_count'] = len(record.x_account_ids)
...
+        <field name="name">x_sequence</field>
+    </record>
+    <record id="field_distribution_key_base_on" model="ir.model.fields">
+        <field name="ttype">selection</field>
+        <field name="copied" eval="True"/>
+        <field name="field_description">Based on</field>
+        <field name="model_id" ref="model_x_distribution_key"/>
+        <field name="name">x_based_on</field>
+    </record>
+    <record id="field_distribution_key_base_meter" model="ir.model.fields">
```
vava-odoo, line 538: since you restructured the file: maybe a pitty to put this one between a field and its field selection values

condominium/data/ir_actions_server.xml
```diff
@@ -17,65 +17,6 @@ for owner in env['res.partner'].search([('x_companies', '=', record.company_id.i
...
-        for ratio in record.x_distribution_key.x_ratio_ids.filtered(lambda l: l.x_ratio != 0):
-            for account in ratio.x_property_ratios.x_account_ids:
-                if record.x_period_start and record.x_period_end:
-                    if account.x_start_date <= record.x_period_end and (not account.x_end_date or record.x_period_start <= account.x_end_date):
-                        # compute overlap start once
-                        overlap_start = max(account.x_start_date, record.x_period_start)
-                        # compute overlap end once
-                        overlap_end = account.x_end_date if account.x_end_date else record.x_period_end
-                        analytic_distribution[account.id] = ((overlap_end - overlap_start).days + 1) * ratio.x_ratio
-                elif account.partner_id.id == ratio.x_owner.id: analytic_distribution[account.id] = ratio.x_ratio
```
vava-odoo, line 52: I don't remember what it was for exactly, but this case is not taken into account in your new action

property_assets_distribution/data/ir_actions_server.xml
```diff
@@ -44,4 +44,52 @@ for line in records:
...
+                if not ratio.x_ratio: continue
+                for account in ratio.x_property_ratios.x_account_ids:
+                    acc_start, acc_end = account.x_start_date, account.x_end_date
+                    if acc_start <= period_end and (not acc_end or period_start <= acc_end):
+                        overlap_start, overlap_end = max(acc_start, period_start), min(acc_end or period_end, period_end)
+                        analytic_distribution[account.id] = analytic_distribution.get(account.id, 0) + ((overlap_end - overlap_start).days + 1) * ratio.x_ratio
+        # Meter distribution
+        elif dist_key.x_based_on == "Meter Readings" and period_start:
+            properties = record.invoice_line_ids.x_property_id or env['x_property'].search([('x_company_id', '=', dist_key.x_company_id.id)])
+            for property in properties:
```
vava-odoo: nit: save one line

property_assets_distribution/data/ir_actions_server.xml
```diff
@@ -44,4 +44,52 @@ for line in records:
...
+                readings = property.x_meter_reading_ids.filtered(lambda mr: mr.x_meter_id.id == dist_key.x_meter.id)
+                if len(readings) < 2: continue
+                for account in property.x_account_ids:
```
vava-odoo, line 81: maybe for later but wouldn't it make sense to archive (or filter out) past accounts?

property_assets_distribution/data/ir_actions_server.xml
```diff
@@ -44,4 +44,52 @@ for line in records:
...
+                    acc_start, acc_end = account.x_start_date, account.x_end_date
+                    if not acc_start or acc_start > period_end or (acc_end and acc_end <= period_start): continue
+                    mrostart = findClosestMeterReading(readings, acc_start)
+                    mroend = findClosestMeterReading(readings, acc_end)
+                    if mrostart.x_date < mroend.x_date:
+                        analytic_distribution[account.id] = abs(mroend.x_quantity - mrostart.x_quantity)
```
vava-odoo, line 87: why not a += here as well?
```suggestion
                        analytic_distribution[account.id] = analytic_distribution.get(account.id, 0) + abs(mroend.x_quantity - mrostart.x_quantity)
```

condominium/data/ir_actions_server.xml
```diff
@@ -17,65 +17,6 @@ for owner in env['res.partner'].search([('x_companies', '=', record.company_id.i
...
-            meter_reading_ids = property.x_meter_reading_ids.filtered(lambda mr: mr.x_meter_id.id == record.x_distribution_key.x_meter.id)
-            if len(meter_reading_ids) >= 2:
-                billstart = findClosestMeterReading(record.x_period_start, meter_reading_ids)
-                billend = findClosestMeterReading(record.x_period_end, meter_reading_ids)
-                for account in property.x_account_ids:
-                    if not account.x_start_date or account.x_start_date > record.x_period_end or (account.x_end_date and account.x_end_date <= record.x_period_start): continue
-                    mrostart = findClosestMeterReading(account.x_start_date, meter_reading_ids)
-                    mrostart_date = mrostart.x_date
-                    billend_date = billend.x_date
-                    if mrostart_date < billstart.x_date: mrostart = billstart
```
vava-odoo, line 65: why did you remove this logic here?

vava-odoo (commented):
You should probably retarget to master (or 19.4 at least if *really* needed)

condominium/data/ir_actions_server.xml
abdrahmanrashed: the if for this elif executes when both `record.x_period_start` and `record.x_period_end` exist. since `record.x_period_start` is required it always exists, meaning we enter the elif block iff `record.x_period_end=False`, if you look closely the entire block is inside an if block that checks for `periods_valid` which is true iff both dates exist.

In otherwords this elif is deadcode that never really runs so there is no reason to keep it.

property_assets_distribution/data/ir_actions_server.xml
abdrahmanrashed: because in python `=` inserts keys in the map if they dont exist but `+=`, `-=`, `*=`, ... dont so it will throw an error

property_assets_distribution/data/ir_actions_server.xml
abdrahmanrashed: filtering out would be a bad idea, we could invoice someone a short while after they leave when the meter bill comes. for archiving maybe we could add something for it but i think we should leave it for @Pierre-savp to decide.

condominium/data/ir_actions_server.xml
abdrahmanrashed: for each account we invoice from the closest meter reading to the start till the closes meter reading to the end, since all these account periods are within the record period any meter reading queries for any account will be within the meter reading taken for the account either way.

also according to @Pierre-savp it is the landlords responsibility to take readings between tenants switching so there should be no conflicts between accounts and bills

condominium/data/ir_rule.xml
abdrahmanrashed: multicompany is a condominium only feature

property_assets_distribution/data/ir_ui_view.xml
abdrahmanrashed: no it is optional, and we invoice based on latest reading

condominium/data/ir_rule.xml
vava-odoo: - install property_assets_distribution
- create a second company

What about the ratios model? Shared?

property_assets_distribution/data/ir_actions_server.xml
vava-odoo: I know that, I pointed out that in the other loop you added previous value if it exists `= analytic_distribution.get(account.id, 0) +`

vava-odoo (approved):
@robodoo r+

vava-odoo:
@robodoo override=ci/industry

PR #2764, opened 2026-08-18
[FIX] .*: fix deposit product configuration

aycha-odoo:
Deposit products are used to account for returned containers, so their
sale price must not be included in the invoice amount. The deposit amount
is tracked separately through the deposit management module.

* Set the Bottle 33cl sale price to 0 in `micro_brewery`.
* Set the Deposit Keg sale price to 0 in `micro_brewery`.
* Keep the deposit amount on the deposit products separately from their
  sale price.
* Configure deposit products with delivery-based invoicing.
* Mark deposit products as storable products so returned deposits can be
  properly handled in stock operations.
* Apply the same configuration to the shared deposit product template used
  by dependent industries.
* Bump the demo module versions to reflect the updated product data.

This ensures returned deposit products can be invoiced without increasing
the invoice amount by their deposit value while retaining the deposit
amount required for return handling.

Task-6469818

### DIFF
--- a/beverage_distributor/__manifest__.py
+++ b/beverage_distributor/__manifest__.py
@@ -1,6 +1,6 @@
 {
     'name': 'Beverage Distributor',
-    'version': '2.4',
+    'version': '2.5',
     'category': 'Supply Chain',
     'depends': [
         'base_automation',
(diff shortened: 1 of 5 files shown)


pgu-odoo:
@vava-odoo Ready to review.
@aycha-odoo Please improve the commit message.

dhrs-odoo:
@aycha-odoo Also instead of using various use .* when you are making changes in multiple industries

vava-odoo (approved):
@robodoo r+

vava-odoo:
@robodoo override=ci/industry

PR #2777, opened 2026-08-19
 [FIX] odoo_partner, photography: repair PDF cross-reference tables and compress assets

aycha-odoo:

PR #2778, opened 2026-08-19
[FIX] odoo_partner, photography: repair PDF cross-reference tables

aycha-odoo:
Fixes noisy `pypdf._reader` log warnings on Runbot: `WARNING pypdf._reader: Ignoring wrong pointing object X 0 (offset 0)`

Static demo PDF files in `odoo_partner` and `photography` had invalid object offset pointers (offset 0) in their xref index tables. When `pypdf` opened these files during test execution, it logged warnings for each bad object lookup.

Re-saved affected PDF assets to rebuild their cross-reference (`xref`) tables with correct object byte locations, eliminating all `pypdf` log warnings on Runbot.


odoo_partner/static/pdf/Agriculture.pdf
vava-odoo, line 1: Good that you found the problem, but the new pdf has now other issues 🥲 

Image

vava-odoo:
Fixed in https://github.com/odoo/industry/pull/2777

PR #2803, opened 2026-08-20
[FIX] construction_developer: fix qweb xpath missing div

trcazier:
Following this PR, https://github.com/odoo/odoo/pull/264918
the template inheritence broke because it was heavily rewritten. This commit fixes this by using the customer_info section instead of the div that existed previously.


vava-odoo:
@robodoo delegate+

trcazier:
@robodoo r+

vava-odoo:
@robodoo r+

vava-odoo:
@robodoo override=ci/industry

PR #2814, opened 2026-08-21
[FIX] construction: fix re-invoice policy for products

dhrs-odoo:
In this PR, we remove the reinvoice_policy field from the construction products since we want to set it as no and its the default value

Task-6486772


vava-odoo:
@robodoo r+

vava-odoo:
@robodoo override=ci/industry

PR #2828, opened 2026-08-21
[FIX] campsite,hotel: restrict synchronization of resources

vava-odoo:
Before this commit, if you create a stay offer resource, an automation would link it to the payment method based on resources. But this is prevented if any pos.session is open. This commit fixes the issue by checking first that there is no open session before adding the resource to the pos.config.


vava-odoo:
@robodoo override=ci/industry

vava-odoo:
ongoing task to remove the automation, closing

PR #2829, opened 2026-08-21
[IMP] hvac_services: update planning and sales access rights for technician

mega-odoo:
- Set `Planning access to User` for technician employees so they can manage their shifts.
- Set `Sales access to All Documents` so technicians can add products to sales orders during planning shifts.

Task ID -  6487833


vava-odoo (approved):
@robodoo r+

vava-odoo:
@robodoo override=ci/industry

PR #2846, opened 2026-08-25
[FIX] .*: remove the url from the tours

dhrs-odoo:
Remove tour URLs from the industry modules to prevent migration failures when migrating from saas-19.2 to saas-19.3.
These URLs are removed starting from 19.0, ensuring they are no longer present in the data during the migration.

OPW - 6494072, 6498705, 6494120


vava-odoo (approved):
@robodoo r+

PR #2853, opened 2026-08-26
[IMP] .*: remove POS resource automation

aycha-odoo:
The automatic assignment of stay resources to POS checkout payment methods can fail when a POS session is already open.

* Remove the base automations and server actions that automatically assign stay resources to the checkout payment method.
* Remove the corresponding data files from the `campsite` and `hotel` manifests.
* Add a warning to the campsite and hotel knowledge articles explaining that resources must be configured on the POS payment methods before opening a POS session.
* Prevent the POS session state from blocking resource configuration.

task-6501591


vava-odoo (approved):
@robodoo r+

PR #2854, opened 2026-08-26
[IMP] real_estate: adapt buyers view

abdrahmanrashed:
Added `x_interested_in` to Buyers menu kanban view creation and set the field to required in Buyers menu kanban and form views

task-6507642

### DIFF
--- a/real_estate/data/ir_ui_view.xml
+++ b/real_estate/data/ir_ui_view.xml
@@ -165,6 +165,19 @@
         </xpath>
     </field>
   </record>
+  <record id="crm_lead_form_inherit" model="ir.ui.view">
+    <field name="name">crm.lead.form.inherit</field>
+    <field name="inherit_id" ref="crm_lead_form_customization_view"/>
+    <field name="mode">primary</field>
+    <field name="active" eval="True"/>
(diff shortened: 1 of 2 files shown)


real_estate/data/ir_ui_view.xml
```diff
@@ -165,6 +165,19 @@
         </xpath>
     </field>
   </record>
+  <record id="crm_lead_form_inherit" model="ir.ui.view">
+    <field name="name">crm.lead.form.inherit</field>
+    <field name="inherit_id" ref="crm_lead_form_customization_view"/>
```
vava-odoo, line 170: couldn't the whole view you inherit from be a primary?

vava-odoo (commented):
missing version bump

real_estate/data/ir_ui_view.xml
abdrahmanrashed: Asked loic about that, and the preexisting customizations should be available from the crm app because we can view buyer leads from there as well as normal leads. the required is added only in the properties app because we create buyers from it, but we can also create normal crm leads from the crm

vava-odoo (approved):
@robodoo r+

PR #2091, opened 2026-04-30
Saas 19.3 fix upgrade vava

vava-odoo:

PR #2868, opened 2026-08-27
[REV] software_reseller: remove demo service_tracking

abdrahmanrashed:
Revert a part of https://github.com/odoo/industry/pull/2091 that overrides the `service_tracking` field in demo due to upgrade script forgeting to check if it is being redefined from data.

task-6048303


vava-odoo (approved):
@robodoo r+

PR #2891, opened 2026-09-01
[FIX] agri_equipment_rental: link stock quants to serial numbers

dhgoy-odoo:
Issue:
- Rental products use Unique Serial Number tracking, but the demo data creates
bulk quants without linking them to specific Serial Numbers.
- The stock is not correctly distributed as one unit per Serial Number.

Solution:
- Replaced bulk quants with one quant per Serial Number (quantity=1) and linked
each to the correct lot_id.

TaskId-6515450


pgu-odoo:

vava-odoo (approved):
@robodoo r+

PR #2915, opened 2026-09-04
[IMP] tests: adapt main company test

frva-odoo:
This commit adapts the test about the main company to
- Allow for any logo and not only the industry one to be present.
- Move the method to be applied in the main test loop, without iterating through the files a second time.

### DIFF
--- a/tests/test_generic/tests/test_xml.py
+++ b/tests/test_generic/tests/test_xml.py
@@ -148,6 +148,7 @@ class TestEnv(IndustryCase):
     def _check_files_in_path(self, module):
         path = get_industry_path() + module
         is_studio_required = False
+        has_main_company_record = False
         static_files = set()
         in_use_files = set()
         checked_records_with_user = {}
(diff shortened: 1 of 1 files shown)


tests/test_generic/tests/test_xml.py
```diff
@@ -213,8 +214,15 @@ def _check_files_in_path(self, module):
                     self._check_base_records_update(tree, file_name, module)
                     if not is_studio_required:
                         is_studio_required = self._check_studio(tree, file_name)
-            if root.split('/')[-1] == 'demo':
-                self._check_main_company_inherit_is_present(root, files, module)
+                if root.split('/')[-1] == 'demo':
+                    has_main_company_record = self._check_main_company_inherit_is_present(tree, file_name, module) or has_main_company_record
```
vava-odoo: invert the order, so that we don't recompute/check if already True (like studio above)

tests/test_generic/tests/test_xml.py
```diff
@@ -213,8 +214,15 @@ def _check_files_in_path(self, module):
...
+
+        if not has_main_company_record and module in self.installed_industries:
```
vava-odoo, line 220: _check_files_in_path is only called on installed modules
```suggestion
        if not has_main_company_record:
```

vava-odoo (approved):

tests/test_generic/tests/test_xml.py
frva-odoo: I put it in the order so that if there was multiple records, we would check the logo for each one. But now I see it makes no sense to have multiple records so I'll switch this :ok_hand:

tests/test_generic/tests/test_xml.py
frva-odoo: There is a difference between installed_modules and installed_industries. Installed_modules contains base_industry_data, booking_engine... while installed_industries is only the industries marked as application. I don't think it makes sense to check for main_company in the technical modules

vava-odoo:
@robodoo r+ delegate+

PR #2936, opened 2026-09-09
[IMP] campsite,hotel: move pos.payment.method to standard

vava-odoo:
Before this commit, the pos.payment.method linked to resources was generated in data module. Since there is now a standard payment method automatically created, this record can be removed from industry modules.

task-6396178


frva-odoo (approved):
Looks good to me 🙂

PR #2942, opened 2026-09-10
[FIX] non_profit_organization: avoid partner edition

vava-odoo:
Before this commit, the server action could edit the partner name and country of an existing partner. This commit modifies it to only create new partner for a new signature.

task-6559796


vava-odoo:
@robodoo up to saas-19.3

vava-odoo:
@robodoo r+

PR #2945, opened 2026-09-10
[FIX] textile_management: fix active settings

vava-odoo:
A recent commit[^1] wrongly introduced a change on activated settings. This commit fixes the issue by activating back right ones.

[^1]: 58a5b145e2bd1f567362ffec9e0b97ed2277b9df


vava-odoo:
@robodoo r+ fw=no

PR #2963, opened 2026-09-14
[FIX] real_estate: trigger automation on stored field

mega-odoo:
Base automation detects when updating stored written fields. When using non-stored fields as `trigger_field_ids` in an automation rule, old values will not be updated due to the following condition: [1] 
`and record._fields[field_name].store`

Link [1]: https://github.com/odoo/odoo/blob/77039cb4f2a386418db83d361973bcae1f238b8f/addons/base_automation/models/base_automation.py#L925

Without old_vals, there are no detected differences, so the `_check_trigger_fields` function returns `False`[2] instead of recognizing the updated field.

Link [2]: https://github.com/odoo/odoo/blob/d25eeaf833fc6fbabd9def2852ecf600f8b7c5ff/addons/base_automation/models/base_automation.py#L878

Since we used `x_is_properties` is a non-stored related field, it does not trigger base automation.

Use `categ_id` field instead, because it is the stored field that is actually updated when the product category is set  properties on a product.

Task ID - 6471332


pgu-odoo:
@vava-odoo Ready to review.

vava-odoo (approved):
@robodoo r+

PR #2966, opened 2026-09-15
[FIX] electronic_refurbishment: Fix margin analysis and demo data

amas-odoo:
- Set the VAT Margin Scheme as both sales and purchase tax on applicable product templates and variants.
- Fix eCommerce product colors, attributes, images, and product variant configuration.
- Ensure the Margin Analysis accounting setting is enabled without relying on demo data.
- Update the default grouping of the product margins report to Published Date: Month, then Product.
- Fix demo stock data to prevent negative stock by validating incoming moves before product conversion and outgoing moves afterward.
- Update the related eCommerce demo data and images.

Task: 6563457


vava-odoo (approved):
@robodoo r+

PR #2985, opened 2026-09-16
[IMP] florist: removed the inventory module and reorder apps

aycha-odoo:
Reorder the main app menus to provide a consistent navigation flow for the florist application and remove the Inventory app, which is not required.

Task-6264296


florist/__manifest__.py
vava-odoo: only the bump and good to go 🙂

vava-odoo (approved):

vava-odoo:
@robodoo r+

PR #2987, opened 2026-09-16
[FIX] booking_engine: remove groupby customer on Schedule

vava-odoo:
A recent commit changes the field definition of
planning.slot.partner_id from a related to a computed. One of the impact of this is that it is now impossible to groupby on this field anymore. This commit therefore removes the groupby filter from the view

https://github.com/odoo/enterprise/commit/9a63146e5373d353f68f58b47d3a4114a308bf0c


vava-odoo:
@robodoo r+

PR #2990, opened 2026-09-16
[FIX] usage_based_maintenance: prevent error when validating a picking

assh-odoo:
Currently, an error occurs when the user validates an internal transfer.

Steps to Reproduce:

 - Install `agri_equipment_rental` industry.
 - Go to `Inventory` > `Products` > `Lots / Serial Numbers`.
 - Create a new `Lot / Serial Number` record and select `Bale Wrapper` as the product.
 - Go to `Inventory` > `Operations` > `Internal` and create a `new internal transfer`.
 - Set `WH/Maintenance` as the destination location and add a line for the `Bale Wrapper` product 
   with a demand of `1`.
 - Click `Mark as Todo`, then set the` quantity to 1` and select the `created serial number` on 
   the transfer line.
 - Click `Validate`, select `Fail`, and click `Confirm`.
 - Click `Validate` again.
```
ValueError: ValueError("Invalid field 'under_warranty' in 'project.task'") while evaluating..........
```
With [commit], industry_fsm_sale was removed and the Field Service functionality was migrated into 
the Planning app,the under_warranty field was also removed.
When the user validates the internal transfer, the base automation runs the server action when its 
domain matches. If the related quality check has failed, the server action creates a corrective task 
for the corresponding lot/serial number. However, the server action still attempts to set the removed 
under_warranty field on the task [1], which raises the error.

This commit removes the under_warranty field from the task creation values to prevent the error.

[commit]: https://github.com/odoo/enterprise/commit/0ce4912d2c76634c2c1249a017ff7c9ab3035725#diff-505826809b4301767ac93163dcfc8e5a148d197f2100cf6f1bb7e0aface0681bL43

[1]- https://github.com/odoo/industry/blob/e4cd034482122158aa8d8e3ad915d26744fa439a/usage_based_maintenance/data/ir_actions_server.xml#L62-L69

sentry-7732795236


maan-odoo (approved):
LGTM

vava-odoo (approved):
@robodoo r+

PR #3000, opened 2026-09-17
[FIX] florist: remove duplicate contact button

aycha-odoo:
The homepage displays two buttons that redirect visitors to the same contact page, making one of them redundant.

Remove the duplicate button and keep the remaining button centered to preserve the intended layout and avoid unnecessary UI elements.

Task-6577992


vava-odoo:
@robodoo r+

PR #3001, opened 2026-09-17
[FIX] *: Master fix errors 0926 - 2

frva-odoo:


vineyard/__manifest__.py
```diff
@@ -1,6 +1,6 @@
 {
     'name': 'Vineyard',
-    'version': '1.5',
+    'version': '1.7',
```
vava-odoo: 👀

food_distribution/data/report_paperformat.xml
vava-odoo, line 19: don't we need to remove this one?

tests/test_booking_engine/tests/test_action_server.py
```diff
@@ -20,7 +20,7 @@ def setUpClass(cls):
         cls.resource = cls.env['resource.resource'].create({
             'name': 'Room Resource Due Out',
             'resource_type': 'material',
-            'calendar_id': False,
+            'calendar_id': cls.env.ref('booking_engine.calendar_1').id,
```
vava-odoo, line 23: @abdrahmanrashed can you validate the commit about test_booking_engine? You worked a lot on calendar issues

abdrahmanrashed: @frva-odoo what happens when you set calendar to all resources to `False`?

frva-odoo: The field is now a required field https://github.com/odoo/odoo/commit/c2a13aa

abdrahmanrashed: i see, since the way fully flexible is defined has changed, don't you think it would be better if we adapt our calendars and set them with `type='undefined'`, `hours_per_week=0`, and `hours_per_day=0` then?

CC: @xavierbol , I think you should see the resource and calendar changes in this commit as well...

abdrahmanrashed (approved):

food_distribution/data/report_paperformat.xml
frva-odoo: I didn't look through it but now that you mention it, it was silently failing as it used the old label template that does not exist anymore. As it was inherited by ours, I put ours instead. And this looks like it is needed as this is not on the same model, we want to be able to print from different places I guess. (was already the case before, inherit was overriding a button on the model I put it on, stock.lot, and there was this new button)

tests/test_booking_engine/tests/test_action_server.py
frva-odoo: This seems like a good idea, I'll change it

vava-odoo (approved):
@robodoo r+ rebase-ff

vava-odoo:
@robodoo override=ci/industry

PR #3002, opened 2026-09-18
[FIX] pet_groomer: fix version key repeatation

amas-odoo:
Fix Dictionary key literal `'version'` repeated


vava-odoo:
@robodoo override=ci/industry  r+ fw=no

PR #3011, opened 2026-09-18
[FIX] beverage_distributor: fix error when importing the module

dhrs-odoo:
When importing beverage_distributor along with its dependencies, `group_stock_multi_locations` was enabled by `beverage_distributor` but not by dependencies such as `excise_management` and `deposit_management`.

As a result, importing the dependencies could attempt to disable the multi-location setting, which fails when the database contains more than one warehouse.

Ensure the setting is handled consistently to avoid the import error.

Task-6577529


pgu-odoo:
@vava-odoo Ready to review.

PR #2977, opened 2026-09-16
[FIX] *: align construction(_developer) module names and images

dhrs-odoo:
Align the module name and image of the Construction
and Construction Developer industries with those used
in the trial database.

Task-6560050

PR #3019, opened 2026-09-18
[FIX] .*: align construction(_developer) module names and images

vava-odoo:
Align the module name and image of the Construction and Construction Developer industries with those used in the trial database.

Task-6560050

backport of odoo/industry#2977


vava-odoo:
@robodoo r+ up to saas-19.3

